PR #10, opened 2018-10-18
add kafka

sfzeng:


sfzeng:
1. Add kafka
2. Test migration from obs to aws with multi-part ability, it work.

PR #32, opened 2018-10-31
Add dist make target for build packaging

stmcginnis:
Add dist make target to be consistent with other OpenSDS projects so
build artifacts can easily be collected in a consistent way across the
different deliverables.


xing-yang:
lgtm

PR #39, opened 2018-11-02
Rename multi-cloud docker images

leonwanghui:
This patch is proposed to remove some invalid characters in the name of docker images, so that we can push them into `docker.io` repository.


wisererik:
LGTM

PR #45, opened 2018-11-05
fix the download bug

liuqinguestc:
fix the download bug

### DIFF
--- a/api/pkg/s3/completemultipartupload.go
+++ b/api/pkg/s3/completemultipartupload.go
@@ -20,7 +20,7 @@ func (s *APIService) CompleteMultipartUpload(request *restful.Request, response 
 	ctx := context.WithValue(request.Request.Context(), "operation", "multipartupload")
 	objectInput := s3.GetObjectInput{Bucket: bucketName, Key: objectKey}
 	objectMD, _ := s.s3Client.GetObject(ctx, &objectInput)
-	//待插入object
+	//to insert object
 	object := s3.Object{}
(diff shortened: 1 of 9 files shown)


api/pkg/s3/completemultipartupload.go
```diff
@@ -29,44 +34,38 @@ func (s *APIService) CompleteMultipartUpload(request *restful.Request, response
...
-
-	objectInput := s3.GetObjectInput{Bucket: bucketName, Key: objectKey}
-	objectMD, _ := s.s3Client.GetObject(ctx, &objectInput)
-
 	if objectMD != nil {
-		_, err := s.s3Client.UpdateObject(ctx, obj)
-		if err != nil {
+		objectMD.Partions=nil
+		objectMD.LastModified = time.Now().String()[:19]
+		//插入元数据
```
leonwanghui: Please use English rather than Chinese.

api/pkg/s3/uploadpart.go
```diff
@@ -15,29 +17,77 @@ import (
 func (s *APIService) UploadPart(request *restful.Request, response *restful.Response) {
 	bucketName := request.PathParameter("bucketName")
 	objectKey := request.PathParameter("objectKey")
-
+	contentLenght := request.HeaderParameter("content-length")
+	size, _ := strconv.ParseInt(contentLenght, 10, 64)
+	//指定backend功能
```
leonwanghui: Same here

PR #63, opened 2018-11-19
add the NoSuchType exception

liuqinguestc:
add the NoSuchType exception


api/pkg/s3/service.go
```diff
@@ -94,7 +94,7 @@ func getBackendClient(s *APIService, bucketName string) datastore.DataStoreAdapt
 	}
 	log.Logf("backendRep is %v:", backendRep)
 	backend := backendRep.Backends[0]
-	client := datastore.Init(backend)
+	client,_ := datastore.Init(backend)
```
leonwanghui: Please use `go fmt` to keep code standard.

PR #64, opened 2018-11-19
1,Support filt object by prefix when migration. 2, Optimize some codes.

sfzeng:
1,Support prefix-filter for migration.
2,Check and make sure plan name is unique. 
3,Optimize logs of datamover and delete unused code.


wisererik:
LGTM

PR #76, opened 2018-11-29
Add Ceph adapter for S3 compatible interfaces

click2cloud-alpha:
This PR adds Ceph adapter for S3 compatible interfaces


leonwanghui:
@click2cloud-alpha Good job, have you tested the e2e feature? And BTW some unit tests may be also required.

api/pkg/s3/datastore/ceph/ceph.go
```diff
@@ -0,0 +1,305 @@
...
+	"encoding/base64"
+	"fmt"
+
+	"io"
+	"io/ioutil"
+	"time"
+
+	"github.com/Click2Cloud/s3client"
+	. "github.com/Click2Cloud/s3client"
+	"github.com/Click2Cloud/s3client/models"
```
xing-yang: Thanks for the PR.  Any reason why these can’t be added to OpenSDS repo directly?

click2cloud-alpha: Do you mean adding this s3client under opensds org?

click2cloud-alpha:
@leonwanghui if you mean testing end to end, then yes we have tested all interfaces except ListParts. Did you mean something else by e2e?
For unit tests, is there any existing location in the repo for unit tests for supported adapters?

leonwanghui:
No, just the same with what you mentioned, and for unit test, you may need to create some test files such like `api/pkg/s3/datastore/ceph/ceph_test.go`

api/pkg/s3/datastore/ceph/ceph.go
xing-yang: Yes

Savasw: @leonwanghui @xing-yang shall i go ahead with changing the repo to opensds in vendor and ceph.go?

click2cloud-alpha: @leonwanghui @xing-yang shall i go ahead with changing the repo to opensds in vendor and ceph.go?

xing-yang:
I see what Leon means.  These are all forked from https://github.com/webrtcn/s3client.  Please use that directly.

api/pkg/s3/datastore/ceph/ceph.go
xing-yang: Please use https://github.com/webrtcn/s3client directly and only add additional changes to OpenSDS.

sfzeng:
LGTM except what Xing and wanghui alread mentioned.

xing-yang:
Thanks for addressing the comments about s3client.  Please add unit tests as Leon suggested.

xing-yang:
I'm merging this now.  Please add unit test in a followup PR.  Thanks.

PR #78, opened 2018-11-30
add bootstrap script for multi-cloud

xxwjj:
Add bootstrap script for multi-cloud, so user can deploy the multi-cloud with one step.


script/bootstrap.sh
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/env bash
+# Copyright (c) 2017 Huawei Technologies Co., Ltd. All Rights Reserved.
```
leonwanghui: Please change the license time to 2018

xxwjj: done

PR #97, opened 2019-01-11
fix backend and bucket bugs

liuqinguestc:
fix backend and bucket bugs


api/pkg/s3/bucketdelete.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/emicklei/go-restful"
 	"github.com/micro/go-log"
 	"github.com/opensds/multi-cloud/api/pkg/policy"
+	. "github.com/opensds/multi-cloud/s3/pkg/exception"
 	"github.com/opensds/multi-cloud/s3/proto"
 	"golang.org/x/net/context"
 	//	"github.com/micro/go-micro/errors"
```
leonwanghui: Can you remove this line if it's not used?

PR #102, opened 2019-02-20
Add GCP adapter for S3 compatible interfaces and added data mover interfaces for GCP backend

Click2Cloud-Gamma:
New updated PR, this PR adds GCP adapter for S3 compatible interfaces and migration scenario from GCP to AWS also tested.

### DIFF
--- a/api/pkg/s3/datastore/gcp/gcp.go
+++ b/api/pkg/s3/datastore/gcp/gcp.go
@@ -1,0 +1,320 @@
+// Copyright (c) 2019 Click2Cloud Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


sfzeng:
Hi, two suggestions:
1. Migration from other cloud  to GCP also need to be tested.
2. Remove duplicated import record, for example, the following two records are duplicate:
"github.com/webrtcn/s3client"
. "github.com/webrtcn/s3client"

Click2Cloud-Gamma:
Hi,
1. Migration from other cloud to GCP is done.
    
2.  "github.com/webrtcn/s3client" 
    . "github.com/webrtcn/s3client" both import have different significant so it generate errors if removed.
(.) import let  the package's exported identifiers  declared in the current file's file block and can be accessed without a qualifier.so some places it is utilising reference.

sfzeng:
Well, for the second one, maybe you can try to reference each identifier exported from the package in the same way, say, with the prefix of (s3client.) 
But that is not necessary, just a suggestion.

Click2Cloud-Gamma:
@sfzeng it is increasing complexity so if current way is accepted please let us know.

sfzeng:
OK, go ahead please.

Click2Cloud-Gamma:
@sfzeng thanks. we are waiting for your final review and merge update.

sfzeng:
@Click2Cloud-Gamma It's OK, need @xing-yang to merge.

wisererik:
LGTM~

PR #114, opened 2019-04-19
Support filter when list objects.

sfzeng:
Support filter by object name prefix and last modified time of objects.
Example:
http://*.*.*.*:***/v1/s3/bucket_name?prefix=a&lastmodified={"gt":"3","lt":"10"}
{"gt":"3","lt":"10"} means objects modified 3 days ago and within 10 days.


api/pkg/common/common.go
```diff
@@ -35,6 +35,8 @@ const (
 	KLimit  = "limit"
 	KOffset = "offset"
 	KSort   = "sort"
+	KLastModified = "lastmodified"
+	KPrefix = "prefix"
```
xxwjj: Filter object by name prfix ,the url request would be like this: /v1/tenantId/bucket?name=like:asw_

sfzeng: it's fixed.

Issue #115, opened 2019-04-19
Speed up the multi-cloud building time.


xxwjj:
Multi-Cloud specified the '-a' flag ,it will force rebuilding of packages that are already up-to-date.  It will take a lot of time.

sfzeng:
It's fixed by this PR #116

PR #116, opened 2019-04-19
Optimize compilation time.

sfzeng:
Optimize the compilation time, now the module without code modification will not be compiled anymore, so the total compile time will decrease extremely most of the time.


xxwjj:
Relative issue: #115

PR #117, opened 2019-04-29
add  docker-compose.yml bali tags

liuqinguestc:
add  docker-compose.yml bali tags

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       - zookeeper
 
   api:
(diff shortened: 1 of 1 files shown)


leonwanghui:
It would still not work because you also need to modify some variables in `opensds-installer`.

leonwanghui:
Resolves [https://github.com/opensds/opensds-installer/issues/143](https://github.com/opensds/opensds-installer/issues/143)

PR #126, opened 2019-05-13
fix auth bug

liuqinguestc:
we need fix the auth bug

the bugs is remained for a long time . it is caused by Himanshu Varshney. I fix the auth bug


leonwanghui:
Please update your comments for this PR, you can refer to this format: This PR is designed(/proposed) to solve {introduce the problem}.

PR #153, opened 2019-06-17
[WIP] Update lifecycle Rule in PUT API

nguptaopensds:
This is the update/modify/edit code in PUT lifecycle

**Release note**:

```release-note
```

### DIFF
--- a/api/pkg/s3/bucketlifecycleput.go
+++ b/api/pkg/s3/bucketlifecycleput.go
@@ -19,6 +19,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
+	"reflect"
 
 	//"sort"
 	"strings"
@@ -116,6 +117,18 @@ func checkValidationOfActions(actions []*s3.Action) error {
 	return nil
 }
 
+func (s *APIService) checkModifyLifecycleConfig(bucketLcCfg *s3.LifecycleRule, rule *s3.LifecycleRule) bool {
+	//check if the rule already exists in lifecycle configuration of the bucket
+	var modifyFlg = false
+
+	res := reflect.DeepEqual(bucketLcCfg, rule)
+	if !res {
+		modifyFlg = true
+		log.Logf("update lifecycle rule is requested")
+	}
+	return modifyFlg
+}
+
 func (s *APIService) BucketLifecyclePut(request *restful.Request, response *restful.Response) {
 	if !policy.Authorize(request, response, "bucket:put") {
 		return
@@ -126,6 +139,9 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
 	bucket, _ := s.s3Client.GetBucket(ctx, &s3.Bucket{Name: bucketName})
 	body := ReadBody(request)
 	log.Logf("MD5 sum for body is %x", md5.Sum(body))
+	modifyFlg := false
+	newbucket := &s3.Bucket{}
+	finalbucket := bucket
 
 	if body != nil {
 		createLifecycleConf := model.LifecycleConfiguration{}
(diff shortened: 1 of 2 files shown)


api/pkg/s3/bucketlifecycleput.go
```diff
@@ -116,6 +117,18 @@ func checkValidationOfActions(actions []*s3.Action) error {
 	return nil
 }
 
+func (s *APIService) checkModifyLifecycleConfig(bucketLcCfg *s3.LifecycleRule, rule *s3.LifecycleRule) bool {
```
rhsakarpos: imho, this is too complicated.
an Update for a lifecycle rule, will simply be a put. from the user pov, 

1. i need to first see the current lifecycle rules on the bucket (we can support a getLifecycleRule(bucket))
2. now, i will update it and call the update/put
3. we support an updateLifecycle(LifecycleRule, bucket). S3 documentation says

> Sets lifecycle configuration for your bucket. If a lifecycle configuration exists, it replaces it.

and we are done !

>

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -214,7 +272,7 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
 
 	// Create bucket with bucket name will check if the bucket exists or not, if it exists
 	// it will internally call UpdateBucket function
-	res, err := s.s3Client.UpdateBucket(ctx, bucket)
```
rhsakarpos: imho, this is incorrect. why are we tampering with the whole bucket here, just to set some new lifecycle rules ?

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -202,8 +218,50 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
...
+							/*
+							get updated bucket from database after deleting the old rule
+							 */
+							newbucket, _ = s.s3Client.GetBucket(ctx, &s3.Bucket{Name: bucketName})
+						} else {
+
+							/*
+							no changes in rule and it is been called again to create
+							 */
+							log.Logf("this rule already exists in configuration : %s\n", rule.ID)
```
rhsakarpos: not needed to compare, check etc.

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -202,8 +218,50 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
...
 
-				// add to the s3 array
-				bucket.LifecycleConfiguration = append(bucket.LifecycleConfiguration, &s3Rule)
+				//check if the there is any modification in rule if it already exists
+				//TODO need to write another modify lifecycle API
+				//Fixme
+				lcCfg := bucket.GetLifecycleConfiguration()
+				for _, cfg := range lcCfg {
+					if cfg.Id == s3Rule.Id {
+						isModifyRequest := s.checkModifyLifecycleConfig(cfg, &s3Rule)
```
rhsakarpos: why do we bother ?

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -202,8 +218,50 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
...
+
+							/*
+								set the modify flag
+							 */
+							modifyFlg = true
+
+							/*
+							Delete previous existing rule from bucket and add new rule with changes
+							 */
+							deleteLcInput := s3.DeleteLifecycleInput{Bucket: bucket.Name, RuleID: cfg.Id}
```
rhsakarpos: ayyyoooo...why are we deleting the rule here ? what if our next update fails ?
just call the below API with the modified rules (that way, the bucket is NEVER without any rules. what if while we are doing all this, some rule timer expires and action has to be taken by S3 ?)
`PutBucketLifecycle`

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -202,8 +218,50 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
...
+							_, err := s.s3Client.DeleteBucketLifecycle(ctx, &deleteLcInput)
```
rhsakarpos: as a guideline, i think we should NEVER delete anything, since the next insert/update may fail, unless we can guarantee a  transaction (with rollback as needed), which we cannot, do not need to

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -214,7 +272,7 @@ func (s *APIService) BucketLifecyclePut(request *restful.Request, response *rest
 
 	// Create bucket with bucket name will check if the bucket exists or not, if it exists
 	// it will internally call UpdateBucket function
-	res, err := s.s3Client.UpdateBucket(ctx, bucket)
+	res, err := s.s3Client.UpdateBucket(ctx, finalbucket)
```
rhsakarpos: we are NOT updating anything on the bucket...why is this API called here ?

api/pkg/s3/bucketlifecycleput.go
```diff
@@ -19,6 +19,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
+	"reflect"
```
rhsakarpos: very expensive, why use it ?

nguptaopensds:
Closing the PR as the context have been changed in new release

PR #172, opened 2019-08-09
Update Progress feature

Click2Cloud-Gamma:
In existing Gelato project the progress calculation was calculated and updated after finishing migration process for each object.

Using this patch progress will increase gradually for migration process.
We have given weightage to each sub-processes of migration which are "Download", "Upload" and "Delete+Completion".
Weightage for Download: 48
Weightage for Upload: 48
Weightage for Delete+Completion: 4

Need to make sure it works with the YIG integration changes
**Release note**:

```release-note
```


Click2Cloud-Gamma:
Hi @xing-yang @wisererik @kumarashit Please review, Thanks.

datamover/pkg/drivers/https/migration.go
```diff
@@ -610,7 +620,7 @@ func runjob(in *pb.RunJobRequest) error {
 					//update database
 					j.PassedCount = (int64(passedCount))
 					j.PassedCapacity = capacity
-					j.Progress = int64(capacity * 100 / j.TotalCapacity)
+					//j.Progress = int64(capacity * 100 / j.TotalCapacity)
 					logger.Printf("capacity:%d,TotalCapacity:%d Progress:%d\n", capacity, j.TotalCapacity, j.Progress)
```
wisererik: Why comment this line?

datamover/pkg/drivers/https/migration.go
```diff
@@ -653,3 +663,19 @@ func runjob(in *pb.RunJobRequest) error {
 
 	return ret
 }
+
+func progress(job *flowtype.Job, size int64, wt float64) {
+	if wt == WT_DOWLOAD {
+		job.MigratedCapacity = job.MigratedCapacity + wt*float64(size)/100
+		job.Progress = int64(job.MigratedCapacity * 100 / float64(job.TotalCapacity))
+		//job.Progress= job.Progress+float64(WT_DOWLOAD*float64(size/job.TotalCapacity))
+		logger.Printf("[INFO] Progress %d", job.Progress)
```
wisererik: remove comment here

dataflow/pkg/model/dataflow_type.go
```diff
@@ -141,15 +141,16 @@ type Job struct {
...
-	SourceLocation string    `json:"sourceLocation" bson:"sourceLocation"`
-	DestLocation   string    `json:"destLocation" bson:"destLocation"`
-	CreateTime     time.Time `json:"createTime" bson:"createTime"`
-	StartTime      time.Time `json:"startTime" bson:"startTime"`
-	EndTime        time.Time `json:"endTime" bson:"endTime"`
-	RemainSource   bool      `json:"remainSource" bson:"remainSource"`
-	Status         string    `json:"status" bson:"status"` //queueing,
-	Tenant         string    `json:"tenant" bson:"tenant"`
-	Progress       int64     `json:"progress" bson:"progress"`
+	MigratedCapacity float64   `json:"migratedCapacity" bson:"migratedCapacity"`
```
kumarashit: This is just a new line for MigratedCapacity added.. Diff show 9 lines..Any whitespaces chage?

datamover/pkg/drivers/https/migration.go
kumarashit: Progress will now be calculated separately. Don't we need to correct the logger statement too here?

datamover/pkg/drivers/https/migration.go
```diff
@@ -653,3 +663,19 @@ func runjob(in *pb.RunJobRequest) error {
 
 	return ret
 }
+
+func progress(job *flowtype.Job, size int64, wt float64) {
```
kumarashit: Put a comment here for the logic used. For readability

datamover/pkg/drivers/https/migration.go
```diff
@@ -653,3 +663,19 @@ func runjob(in *pb.RunJobRequest) error {
...
+	if wt == WT_DOWLOAD {
+		job.MigratedCapacity = job.MigratedCapacity + wt*float64(size)/100
+		job.Progress = int64(job.MigratedCapacity * 100 / float64(job.TotalCapacity))
+		//job.Progress= job.Progress+float64(WT_DOWLOAD*float64(size/job.TotalCapacity))
+		logger.Printf("[INFO] Progress %d", job.Progress)
+		db.DbAdapter.UpdateJob(job)
+	} else {
```
kumarashit: This will be for WT_UPLOAD and WT_DELETE both. Is this valid for both?

datamover/pkg/drivers/https/prepare.go
```diff
@@ -139,11 +139,24 @@ func getConnLocation(ctx context.Context, conn *pb.Connector) (*LocationInfo, er
 
 	return nil, errors.New("unsupport type")
 }
-
 func getObjs(ctx context.Context, in *pb.RunJobRequest, defaultSrcLoca *LocationInfo, offset, limit int32) ([]*osdss3.Object, error) {
+
 	switch in.SourceConn.Type {
 	case flowtype.STOR_TYPE_OPENSDS:
 		return getOsdsS3Objs(ctx, in, offset, limit)
+	case flowtype.STOR_TYPE_AWS_S3:
```
kumarashit: Are these relevant changes for progress?

datamover/pkg/drivers/https/migration.go
Click2Cloud-Gamma: Hi @wisererik . Thanks for the review. As we are calculating Progress code in progress function defined at the end, we will not need to calculate progress here. Thanks

datamover/pkg/drivers/https/prepare.go
Click2Cloud-Gamma: No but when we are using APIs, migration process gives error at the beginning like "unsupported storage type - aws-s3" .
Thanks

kumarashit: Ok. Just for more tracking purpose. IMHO, we should open a new issue for migration and then push these under that, if it doesn't impact progress code.

datamover/pkg/azure/blob/blobmover.go
```diff
@@ -52,6 +52,9 @@ func handleAzureBlobErrors(err error) error {
 			case string(azblob.StorageErrorCodeAuthenticationFailed):
 				log.Log("azure error: permission denied.")
 				return errors.New(DMERR_NoPermission)
+			case string(azblob.StorageErrorCodeContainerNotFound):
+				log.Log("azure error: container not found.")
+				return errors.New(DMERR_BucketNotFound)
```
sfzeng: It seems this error is defined, but not used, so it seems not necessary.

datamover/pkg/drivers/https/prepare.go
sfzeng: Which kind of migration you used? If it is the migration from one OpenSDS bucket to another OpenSDS bucket, then no such error will return.  Actually, those 'unsupported storage type' code is for migration between non OpenSDS bucket which is not supported now. You can not just add codes like this to support it, it need many other works.

datamover/pkg/drivers/lifecycle/crosscloud_transition.go
```diff
@@ -67,11 +68,11 @@ func copyObj(ctx context.Context, obj *osdss3.Object, src *BackendInfo, dest *Ba
 		log.Logf("the transition of object[%s] is in-progress\n", obj.ObjectKey)
 		return errors.New(DMERR_TransitionInprogress)
 	}
-
+	var job *model.Job
```
sfzeng: The total capacity of the whole job is used when you update progress in the function of progress, but it's value is assigned nowhere,  that means there must be 'division by zero' error.

sfzeng: BTW, this file is for lifecycle, not migration, lifecycle do not need progress.

datamover/pkg/drivers/https/prepare.go
Click2Cloud-Gamma: Okay I will remove this part. Thanks

datamover/pkg/drivers/https/migration.go
Click2Cloud-Gamma: Done.. Thanks

datamover/pkg/azure/blob/blobmover.go
Click2Cloud-Gamma: Using this we will get the specific error else it shows big error so this will allow us to detect easily that bucket is not available instead of big error message for Azure. Thanks

datamover/pkg/drivers/lifecycle/crosscloud_transition.go
Click2Cloud-Gamma: Since crosscloud_transition also use same function which is defined for migration, I have added a dummy job here. To handle the "division by zero" I will add some filters in progress calculation to skip the calculation for lifecycle related process.

Click2Cloud-Gamma:
Hi @kumarashit @sfzeng @wisererik. Please review, Thanks

wisererik:
@sfzeng @kumarashit please review it again, it looks fine now.

datamover/pkg/drivers/https/migration.go
```diff
@@ -653,3 +673,14 @@ func runjob(in *pb.RunJobRequest) error {
...
+// To calculate Progress of migration process
+func progress(job *flowtype.Job, size int64, wt float64) {
+	// Migrated Capacity = Old_migrated capacity + WT(Process)*Size of Object/100
+	MigratedCapacity := job.MigratedCapacity + float64(size)*(wt/100)
+	job.MigratedCapacity = math.Round(MigratedCapacity*100) / 100
+	// Progress = Migrated Capacity*100/ Total Capacity
+	job.Progress = int64(job.MigratedCapacity * 100 / float64(job.TotalCapacity))
+	logger.Printf("[INFO] Progress %d", job.Progress)
+	db.DbAdapter.UpdateJob(job)
+}
```
sfzeng: progress is used by multiple go-routines running in parallel, how to make sure it is concurrent security?

datamover/pkg/drivers/https/prepare.go
```diff
@@ -139,8 +139,8 @@ func getConnLocation(ctx context.Context, conn *pb.Connector) (*LocationInfo, er
 
 	return nil, errors.New("unsupport type")
 }
-
```
sfzeng: Please leave the blank line here.

datamover/pkg/drivers/https/prepare.go
```diff
@@ -139,8 +139,8 @@ func getConnLocation(ctx context.Context, conn *pb.Connector) (*LocationInfo, er
...
 func getObjs(ctx context.Context, in *pb.RunJobRequest, defaultSrcLoca *LocationInfo, offset, limit int32) ([]*osdss3.Object, error) {
+
```
sfzeng: Please remove the blank line here.

Click2Cloud-Gamma: Done, Thanks.

datamover/pkg/drivers/https/migration.go
Click2Cloud-Gamma: I have tested this for multiple scenarios( in which go-routines were running in parallel ) , it has worked fine in all cases. Is there any way using which I can check for concurrent security.
Thanks

wisererik:
@Click2Cloud-Gamma, please remove WIP. @sfzeng please review it again about your comments.

datamover/pkg/drivers/https/migration.go
sfzeng: Did you test several objects in a single migrating task, or several migrating tasks running in parallel? IMO, in theory, the concurrent security only exist in the first senario, and the problem is random, maybe you can test with more objects in a single migrating task and check if the progress is increasing all the time and finished with 100% at last.

Click2Cloud-Gamma: Yes we have tested both scenario, 1. Single migrating task with several objects and Multiple jobs running in parallel. In both scenario the progress was increasing all the time during migration and it finished with 100% at the end of migration.

sfzeng: Hi, I did a test and found something wrong, details is as below, please check.
# Branch lick2Cloud/multi-cloud > multi_cloud_feature
## Source bucket
Totally 42 objects with the same size, as below:
source bucket1
source bucket2
## Result
** What get from the dashboard
dashboard-1
** What get from database
db-1
** What get from log
log-1

# Branch opensds/multi-cloud > master
## Source bucket
It's the same as above.
## Result:
** What get from the dashboard
dashboard-2
** What get from database
db-2
** What get from log
log-2

Click2Cloud-Gamma: @sfzeng Thanks for your time and testing this feature. By mistake `Passedcapacity` got deleted while adding commits. Now I have solved this issue. I am testing it again. Thanks

datamover/pkg/drivers/lifecycle/crosscloud_transition.go
sfzeng: Thanks for your update, I tested it follow the steps I used before, and it seems it works, and when I checked the log, I found those routines rarely call progress() at the same time, so I made a simulation test, the step is as below:
1. Change simuRoutines to be 60;
2. Comment out those codes that used to upload and download from clouds in the function of MoveObj to simulate that several goroutines calling the function of progress() simultaneously.
And I got the logs as below:
log
The log shows before uploading, the progress already reached 91%, from the point of my view, it's not reasonable.

P.S.  I cannot access google drive currently, so if you need the test code and logs, please give me an email, I'll send to you. Thanks.

sfzeng: And for the problem, that big object cannot upload, I'll check it later.

sfzeng: Hi, for the crash you showed above, I think it's because bad abnormal handling, I created an issue [#183 ](https://github.com/opensds/multi-cloud/issues/183), and a PR [#184](https://github.com/opensds/multi-cloud/pull/184) to fix it.
And I think you probably encountered some environment exceptions, cause I can upload object with multipart upload successfully, even without PR#184:
dashboard - 1
log - 1

wisererik:
How about this PR now?

datamover/pkg/drivers/lifecycle/crosscloud_transition.go
Click2Cloud-Gamma: > Thanks for your update, I tested it follow the steps I used before, and it seems it works, and when I checked the log, I found those routines rarely call progress() at the same time, so I made a simulation test, the step is as below:
> 
> 1. Change simuRoutines to be 60;
> 2. Comment out those codes that used to upload and download from clouds in the function of MoveObj to simulate that several goroutines calling the function of progress() simultaneously.
>    And I got the logs as below:
>    log
>    The log shows before uploading, the progress already reached 91%, from the point of my view, it's not reasonable.
> 
> P.S. I cannot access google drive currently, so if you need the test code and logs, please give me an email, I'll send to you. Thanks.

@sfzeng Please share full logs if possible to roshan.wanaskar@click2cloud.net

sfzeng: Hi @Click2Cloud-Gamma, sorry for late reply, the log is sent, please check. Thanks.

sfzeng:
From the point of my view, there maybe a concurrency problem, logs have been sent to roshan.wanaskar@click2cloud.net, please check, thanks.

sfzeng:
LGTM

PR #175, opened 2019-08-23
Delete Objects after Migrating all objects

Click2Cloud-Gamma:
In current OpenSDS, when we migrate objects for eg. 10 objects (remain source is false), when each objects gets migrated successfully it gets deleted before migration of other objects.

In case of failure eg. if 3 objects are already migrated, it will get deleted too from source and if migration got failed then we will have 7 objects in source (3 objects got deleted) and 3 objects in destination. 

Using this patch, in same scenario of migration failure, after failure we will have 10 objects in source (No loss of objects from Source bucket) and destination will have 3 objects. So at least at source bucket user will get access to all objects after failure also.

### DIFF
--- a/datamover/pkg/drivers/https/migration.go
+++ b/datamover/pkg/drivers/https/migration.go
@@ -80,7 +80,7 @@ func HandleMsg(msgData []byte) error {
 }
 
 func doMove(ctx context.Context, objs []*osdss3.Object, capa chan int64, th chan int, srcLoca *LocationInfo,
-	destLoca *LocationInfo, remainSource bool) {
+	destLoca *LocationInfo) {
 	//Only three routines allowed to be running at the same time
 	//th := make(chan int, simuRoutines)
 	locMap := make(map[string]*LocationInfo)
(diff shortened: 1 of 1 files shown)


Click2Cloud-Gamma:
Hi @kumarashit @sfzeng @wisererik. Please review, Thanks

wisererik:
if the 3 objects are very big, it's costly to keep object in source bucket and target bucket.

kumarashit:
IMO, This should be user discretion to either keep it or delete it. There is cost involved here so user can take a call

sfzeng:
Before migration finished, the user needs to pay double the price for those migrated objects. So IMO,  it's better to give the user to make a choice or leave it as it now.

Click2Cloud-Gamma:
> Before migration finished, the user needs to pay double the price for those migrated objects. So IMO, it's better to give the user to make a choice or leave it as it now.

So I think the existing "remainSource" parameter is enough for that, no need to add 1 more parameter.

sfzeng:
> > Before migration finished, the user needs to pay double the price for those migrated objects. So IMO, it's better to give the user to make a choice or leave it as it now.
> 
> So I think the existing "remainSource" parameter is enough for that, no need to add 1 more parameter.

But does this PR is talking about, if user choose to not remain source, then when will we delete the source object?

Click2Cloud-Gamma:
> > > Before migration finished, the user needs to pay double the price for those migrated objects. So IMO, it's better to give the user to make a choice or leave it as it now.
> > 
> > 
> > So I think the existing "remainSource" parameter is enough for that, no need to add 1 more parameter.
> 
> But does this PR is talking about, if user choose to not remain source, then when will we delete the source object?

Using this patch, If `remainSource=false` ,all the objects will get deleted after migrating ALL objects.

PR #684, opened 2019-10-29
add global id generator according to snowflake algo.

hopkings2008:
This PR adds a global id generator for yig storage driver. Because when create a multipart, we need a global unique id for the multipart upload, we can use that global id generator to generate this kind of 64-bit id.
The implementation of global id generator refers to the snowflake algo.
Also, tests are used to cover the function of global id generator.

**Release note**:

```release-note
```

### DIFF
--- a/s3/pkg/datastore/yig/tests/utils_test.go
+++ b/s3/pkg/datastore/yig/tests/utils_test.go
@@ -1,0 +1,56 @@
+package tests
+
+import (
+	"sync"
+
+	"github.com/opensds/multi-cloud/s3/pkg/datastore/yig/utils"
+	. "gopkg.in/check.v1"
+)
+
+func (ys *YigSuite) TestGlobalId(c *C) {
+	gi, err := utils.NewGlobalIdGen()
+	c.Assert(err, Equals, nil)
+	c.Assert(gi, Not(Equals), nil)
+	ids := make(map[int64]int64)
+	count := 8192
+
+	for i := 0; i < count; i++ {
+		id := gi.GetId()
+		ids[id] = id
+	}
+
+	c.Assert(len(ids), Equals, count)
+}
+
+func (ys *YigSuite) TestGlobalIdConcurrent(c *C) {
+	var wg sync.WaitGroup
+	gi, _ := utils.NewGlobalIdGen()
+	ids := make(map[int64]int64)
+	count := 8192
+	numThreads := 100
+	idsChan := make(chan int64)
+	wg.Add(numThreads)
+
+	funcGen := func(loop int) {
+		for i := 0; i < loop; i++ {
+			id := gi.GetId()
+			idsChan <- id
+		}
+		wg.Done()
+	}
+
+	for i := 0; i < numThreads; i++ {
+		go funcGen(count)
+	}
+
+	go func() {
+		wg.Wait()
+		close(idsChan)
+	}()
+
+	for id := range idsChan {
+		ids[id] = id
+	}
+
+	c.Assert(len(ids), Equals, numThreads*count)
+}
--- a/s3/pkg/datastore/yig/tests/utils_test.go
+++ b/s3/pkg/datastore/yig/tests/utils_test.go
@@ -1,3 +1,17 @@
+// Copyright 2019 The OpenSDS Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package tests
 
 import (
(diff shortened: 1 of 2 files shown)


hopkings2008:
@rhsakarpos @nguptaopensds please help to take a look at this PR, this PR provides a global id generator according to snowflake algo, and this util is used by yig storage driver, thanks a lot.

s3/pkg/datastore/yig/tests/utils_test.go
```diff
@@ -0,0 +1,56 @@
...
+	. "gopkg.in/check.v1"
+)
+
+func (ys *YigSuite) TestGlobalId(c *C) {
+	gi, err := utils.NewGlobalIdGen()
+	c.Assert(err, Equals, nil)
+	c.Assert(gi, Not(Equals), nil)
+	ids := make(map[int64]int64)
+	count := 8192
+
```
wisererik: why it's 8192, please define a constant with a description

s3/pkg/datastore/yig/utils/global_id.go
```diff
@@ -0,0 +1,74 @@
...
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return 0, err
+	}
+	for _, address := range addrs {
+		// check the address type and if it is not a loopback the display it
+		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			ip := ipnet.IP.To4()
+			if ip != nil && (ip[0] == 10 || ip[0] == 172 && (ip[1] >= 16 && ip[1] < 32) || ip[0] == 192 && ip[1] == 168) {
+				return (int64(ip[2])<<8 + int64(ip[3])) & 0x0fff, nil
```
wisererik: is there any other method to validate loopback address?

s3/pkg/datastore/yig/tests/utils_test.go
hopkings2008: thanks @wisererik for your suggestion, 8192 is just a number which contains boundary for the max number of ids per ms. i will define a constant for it.

s3/pkg/datastore/yig/tests/utils_test.go
```diff
@@ -0,0 +1,56 @@
+package tests
```
sfzeng: Hi @hopkings2008  Please add a license announcement at the beginning. At the same time, add some notes to explain it used the snowflake. Thanks

hopkings2008: ok, very good suggestion @sfzeng , i will amend it according to your suggestion.

s3/pkg/datastore/yig/utils/global_id.go
```diff
@@ -0,0 +1,74 @@
+package utils
```
rhsakarpos: can we use this instead ? BSD2 license, looks good to me
https://github.com/bwmarrin/snowflake

sfzeng: Hi @hopkings2008  Please add a license announcement at the beginning. At the same time, add some notes to explain it used the snowflake. Thanks.

s3/pkg/datastore/yig/tests/utils_test.go
```diff
@@ -0,0 +1,56 @@
...
+	go func() {
+		wg.Wait()
+		close(idsChan)
+	}()
+
+	for id := range idsChan {
+		ids[id] = id
+	}
+
+	c.Assert(len(ids), Equals, numThreads*count)
```
rhsakarpos: can you please point me to the check that confirms that all the ids are unique ?

s3/pkg/datastore/yig/utils/global_id.go
hopkings2008: @wisererik i need make a investigation about how to check the ip address. this kind of check is used for the 'machine id' in snowflake algo. by default, to make it simple to use, i omit the requirement to input the 'machine id'. i will check whether there is another good way to query the ip. thanks a lot.

s3/pkg/datastore/yig/tests/utils_test.go
rhsakarpos: here, it could have generated n ids for n threads, not necessarily unique, right ?

s3/pkg/datastore/yig/utils/global_id.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
```
rhsakarpos: since the idea is to generate global unique ids, should we make this a micro-service ?

s3/pkg/datastore/yig/utils/global_id.go
```diff
@@ -0,0 +1,74 @@
...
+func (gi *GlobalIdGen) getMachineId() (int64, error) {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return 0, err
+	}
+	for _, address := range addrs {
+		// check the address type and if it is not a loopback the display it
+		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			ip := ipnet.IP.To4()
+			if ip != nil && (ip[0] == 10 || ip[0] == 172 && (ip[1] >= 16 && ip[1] < 32) || ip[0] == 192 && ip[1] == 168) {
```
rhsakarpos: what is 10, 172, 192 etc. to compare ip[0] ? the intent of this code, please add comment as to what we want to to here, thanks

s3/pkg/datastore/yig/utils/global_id.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	id, err := gi.getMachineId()
+	if err != nil {
+		return nil, err
+	}
+	gi.MachineId = id
+	return gi, nil
+}
+
+func (gi *GlobalIdGen) GetId() int64 {
```
rhsakarpos: question...will a running, incrementing sequence number suffice ?  much simpler to handle imho

s3/pkg/datastore/yig/utils/global_id.go
hopkings2008: @rhsakarpos thanks for your suggestion, i will investigate about this kind of check according to @wisererik 's suggestion, and also, i will make a enhancement according to your suggestion, thanks

s3/pkg/datastore/yig/utils/global_id.go
hopkings2008: thanks @sfzeng for your suggestion, i will add that.

hopkings2008: thanks @rhsakarpos for your suggestion, please trust me that i have searched the similar implementation before, but the reason for why i implement it directly is that the algo is simple itself and we don't need to depends on the new 3rd lib, and also, we can customize it for that we may not need to transfer a 'machine id' when use it.

s3/pkg/datastore/yig/utils/global_id.go
hopkings2008: @rhsakarpos because according to snowflake algo, the global unique id has a 'machine id' in it. So, the algo will generate the ids monotonically per machine.

s3/pkg/datastore/yig/tests/utils_test.go
hopkings2008: @rhsakarpos this test will check wether the implementation will generate n unique ids for n threads. And because i use map to store the ids, so if there are ids which are the same, len(ids) will be smaller than the n * n number.

s3/pkg/datastore/yig/utils/global_id.go
hopkings2008: @rhsakarpos currently, yig storage driver needs this kind of function to generate the global unique id for multipart upload id. if other modules also need that function, we can discuss to make it as a micro-service, thanks.

rhsakarpos:
thank you for your patient responses, LGTM

hopkings2008:
@wisererik i have made a change for retrieving ipv4 address, please help to review it again.
@rhsakarpos @sfzeng I have added one optional input parameter when get GlobalIdGen, it is machine id. if the value of it is less than or equal to 0, we will use the lower bits of ipv4 address for the machine id. and please review this, and welcome your suggestions, thanks.

sfzeng:
LGTM

PR #688, opened 2019-11-04
Support get bucket (list obejcts in a bucket)

sfzeng:
This PR is to support the API of  **GET bucket**, which is used to list objects in a specific bucket. It is compatible with the standard S3 API.

When you use the API of **Get bucket**, you can specify some parameters, like max-keys, prefix, start-after, just the same as standard AWS s3 API support. Besides that, we also support to filter objects by its last modified time and its storage tier, which will be used by lifecycle management.

**Release note**:

```
```


sfzeng:
Closed it because I choose the wrong target branch, should be the development branch, not the master branch.

PR #704, opened 2019-11-06
fix the problem of panicing the process if there is no ceph conf…

hopkings2008:
… files for yig storage driver.

this pr fixes the problem of panicing the process if there is no ceph conf files.
when multi-cloud s3 starts, it will try to load the storage drivers such as yig driver, but if there is no ceph conf files, yig driver will panic the process which make s3 down, in this case, the create bucket will fail.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # 701

**Release note**:

```release-note
```


hopkings2008:
@rhsakarpos @nguptaopensds pls help to review this pr, and this pr fix the problem of fail to create bucket on aws. thanks a lot.

PR #709, opened 2019-11-07
add db operation functions in yig storage backend and move cluster related functions from service layer to yig storage backend.

hopkings2008:
Because cluster info contains ceph related information and it is only used by yig storage backend, this pr moves the cluster info related functions from service layer to yig storage backend.

**Release note**:

```release-note
```


sfzeng:
Hi @hopkings2008 , does yig storage backend only use database to save and get cluster information? Is this database seperated from s3 database?

hopkings2008:
> Hi @hopkings2008 , does yig storage backend only use database to save and get cluster information? Is this database seperated from s3 database?

@sfzeng yig storage backend will save the cluster info and multipart related info. and the multipart info only contains 'storage information' and it doesn't contain the upper layer information such as uid , bucketname, objectname, etc. So we need to split these two logic data model. and the backend driver needs to separate the db from the s3 db, since they are two different layers.
what do you think of it? and welcome your suggestions.

s3/pkg/meta/db/db.go
```diff
@@ -46,8 +46,6 @@ type DBAdapter interface {
 	//PutObjectMap(objMap *ObjMap, tx interface{}) error
 	//DeleteObjectMap(objMap *ObjMap, tx interface{}) error
 
-	//cluster
-	GetCluster(fsid, pool string) (cluster Cluster, err error)
 	//lc
 	/*
```
wisererik: please remove unused code

hopkings2008: @wisererik thanks for your review, and the unused code has been removed.

s3/pkg/datastore/yig/conf/yig.toml
```diff
@@ -9,4 +9,5 @@ url="default"
 ceph_dir="/etc/ceph/*.conf"
 
 [database]
+db_type="tidb"
 db_url="root:@tcp(tidb:4000)/yig"
```
rhsakarpos: it is highly unsafe to have a root user without a password and use it in application. please do
1. run the mysql secure install script (hardening the mysql)
2. create an application user on the DB, with priveleges granted to what is needed by the application
3. use that user and credentials in our connection code

s3/pkg/datastore/yig/config/config.go
```diff
@@ -135,21 +135,32 @@ func (cc *CacheConfig) Parse(vals map[string]interface{}) error {
...
+	DbType       string
+	DbUrl        string
+	DbPassword   string
+	MaxIdleConns int
+	MaxOpenConns int
 }
 
 func (dc *DatabaseConfig) Parse(vals map[string]interface{}) error {
+	dc.MaxIdleConns = 0
+	dc.MaxOpenConns = 1024
```
rhsakarpos: why so many ?

s3/pkg/datastore/yig/config/config.go
```diff
@@ -135,21 +135,32 @@ func (cc *CacheConfig) Parse(vals map[string]interface{}) error {
...
-	DbPassword string
+	DbType       string
+	DbUrl        string
+	DbPassword   string
+	MaxIdleConns int
+	MaxOpenConns int
 }
 
 func (dc *DatabaseConfig) Parse(vals map[string]interface{}) error {
+	dc.MaxIdleConns = 0
```
rhsakarpos: why is this 0 ? how will connection pooling work with no idle connections ?

s3/pkg/datastore/yig/meta/db/tidb/cluster.go
```diff
@@ -0,0 +1,16 @@
+package tidb
+
+import (
+	"github.com/opensds/multi-cloud/s3/pkg/datastore/yig/meta/types"
+)
+
+//cluster
```
rhsakarpos: is this YIG CEPH specific ? cluster == backend ? in multicloud case ?

s3/pkg/datastore/yig/meta/db/tidb/driver.go
```diff
@@ -0,0 +1,36 @@
...
+	TIDB_DRIVER_ID = "tidb"
+)
+
+type TidbDriver struct {
+}
+
+func (td *TidbDriver) OpenDB(dbCfg config.DatabaseConfig) (driver.DB, error) {
+	db := &Tidb{}
+	var err error
+	db.DB, err = sql.Open("mysql", dbCfg.DbUrl)
```
rhsakarpos: do we use connection pooling ?

s3/pkg/datastore/yig/meta/db/tidb/driver.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+const (
+	TIDB_DRIVER_ID = "tidb"
+)
+
+type TidbDriver struct {
+}
+
+func (td *TidbDriver) OpenDB(dbCfg config.DatabaseConfig) (driver.DB, error) {
```
rhsakarpos: please ensure this method is idempotent

s3/pkg/datastore/yig/config/config.go
hopkings2008: Thanks @rhsakarpos for your review, because if there is so high work load, the more max open conns  will reduce the latency between the yig storage backend and the tidb.

s3/pkg/datastore/yig/meta/db/tidb/cluster.go
hopkings2008: @rhsakarpos yig backend will manage more than one ceph cluster, so yig backend is a top management layer. So, backend will contain more than one ceph cluster.

s3/pkg/datastore/yig/meta/db/tidb/driver.go
hopkings2008: @rhsakarpos  thanks for your review, and what do you mean the idempotent in this function? and original object of this function is to create a new db handle for caller usage. Each call to openDB will create a new db handle.

s3/pkg/datastore/yig/meta/db/tidb/driver.go
hopkings2008: @rhsakarpos the function call 'db.DB.SetMaxIdleConns' and 'db.DB.SetMaxOpenConns' will work for the connection pooling.

s3/pkg/datastore/yig/config/config.go
hopkings2008: @rhsakarpos thanks for your nicely review, and i have changed the value to a macro which by default is 1024. both max idle conns and max open conns are 1024, and both values are defined in separate macro.

s3/pkg/datastore/yig/conf/yig.toml
hopkings2008: @rhsakarpos thanks for your kindly review. and the yig.toml in our source code is just the sample. and it is only used in ci/test environment. In production environment, the deployment team will handle this problem. and i have link this problem into 'https://github.com/opensds/multi-cloud/issues/683', and add comment in that issue. thanks a lot.

sfzeng: Where do we initiated database yig?

hopkings2008: @sfzeng there is a script which init the tables in db before yig backend  is registered. this change will in next pr.

s3/pkg/datastore/yig/meta/db/tidb/driver.go
rhsakarpos: idempotent will mean, no matter how many times the function is called, create only one DB handle...something like
```
synchronized()
{
    if(dbHandle == null)
    {
        createDBHandle()
    }
}
```
Depends on the need...do we need only one DB handle across the runtime of the application or many

rhsakarpos:
@hopkings2008 thanks for the updates, LGTM

s3/pkg/datastore/yig/meta/db/tidb/driver.go
hopkings2008: @rhsakarpos thanks for your good explanation, and for one yig backend instance, there is only one db handle, and this is controlled by the upper logic, so, in our scenario, yig backend will control to create only one db handle for usage. and for above OpenDB function, it will always return a new db handle when it is called.

s3/pkg/datastore/yig/conf/yig.toml
sfzeng: @hopkings2008 Could you please create an issue for this and comment here, thanks.

hopkings2008: @sfzeng thanks for your review, i have created the issue 'https://github.com/opensds/multi-cloud/issues/714' to track this problem you asked.

PR #717, opened 2019-11-11
use int instead of int64 for PartNumber

hopkings2008:
this PR changes the type of PartNumber from int64 to int, since PartNumber can be only between 1 and 10000 according to https://docs.aws.amazon.com/zh_cn/AmazonS3/latest/dev/qfacts.html

**Release note**:

```release-note
```


hopkings2008:
due to the type changes of PartNumber relates to huawei, aws, and yig backend, just close this pr and submit anther pr which contains the whole change. thanks.

Issue #694, opened 2019-11-05
Support lifecycle management


sfzeng:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
> /kind feature

**Anything else we need to know?**:
Include: 
1. PUT/GET/DELETE lifecycle rules, API should be compatible with AWS S3.
2. Lifecycle management should support expiration, in-cloud transition and cross-cloud transition.
3. Refactor lifecycle management based on COPY-Object API provided by OpenSDS S3 service.

**Environment**:
- Gelato(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

kumarashit:
PR is raised

PR #720, opened 2019-11-12
PUT/GET/DELETE lifecycle configurations

sfzeng:
This PR is to support the APIs of PUT/GET/DELETE lifecycle configurations, lifecycle schedule and data migrations are left behind.

This PR will fix a part of [#694](https://github.com/opensds/multi-cloud/issues/694)

When DELETE lifecycle is called, all lifecycle rules of the bucket will be deleted, this is different from the last version, we changed it to be like this because this is compatible with AWS S3. And if the user wants to delete a specific rule, the PUT API can be used.

**Release note**:

```release-note
```


hopkings2008:
it is fine, LGTM

yuejuSDS:
With respect to changes to lifeCycle's delete function, dashboard also needs to synchronize the changes to ensure that the delete function is used properly.

PR #722, opened 2019-11-13
Add feature to get object

### LINKED ISSUE #721: Support get object

>
> /kind feature

**What happened**:

Support get object

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:

The PR implement the function to get object from s3 service. It also contains http features of range get, and conditional headers.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #721

**Release note**:

```release-note
```


api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+			eof = true
+		}
+
+		if rsp.ErrorCode != int32(s3error.ErrNoErr) {
+			s3err = rsp.ErrorCode
+			log.Errorf("received s3 service error, error code:", rsp.ErrorCode)
+			break
+		}
+		if len(rsp.Data) == 0 {
+			continue
```
hopkings2008: it is good to break intead  of continue here. and if len(rsp.Data) =0, err should be io.EOF, if not, it is a bug of under storage layer.

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+	eof := false
+	left := length
+	for !eof && left > 0 {
+		rsp, err := stream.Recv()
+		if err != nil && err != io.EOF {
+			log.Errorln("recv err", err)
+			break
+		}
+		if err == io.EOF {
+			eof = true
```
hopkings2008: it is good to break directly here.

api/pkg/s3/api-headers.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/opensds/multi-cloud/s3/pkg/meta/types"
+	pb "github.com/opensds/multi-cloud/s3/proto"
+)
+
+// Refer: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html
+var CommonS3ResponseHeaders = []string{"Content-Length", "Content-Type", "Connection", "Date", "ETag", "Server",
+	"x-amz-delete-marker", "x-amz-id-2", "x-amz-request-id", "x-amz-version-id"}
+
+// Encodes the response headers into XML format.
+func EncodeResponse(response interface{}) []byte {
```
sfzeng: It seems this function is duplicate with the one in api-response.go

api/pkg/s3/api-headers.go
```diff
@@ -0,0 +1,77 @@
...
+// Write object header
+func SetObjectHeaders(response *restful.Response, object *pb.Object, contentRange *HttpRange) {
+	// set object-related metadata headers
+	w := response.ResponseWriter
+	lastModified := time.Unix(object.LastModified, 0).UTC().Format(http.TimeFormat)
+	response.ResponseWriter.Header().Set("Last-Modified", lastModified)
+
+	w.Header().Set("Content-Type", object.ContentType)
+	if object.Etag != "" {
+		//	w.Header().Set("ETag", "\""+object.Etag+"\"")
```
sfzeng: Remove unused code, please.

api/pkg/s3/objectget.go
```diff
@@ -15,7 +15,16 @@
...
+	"net/url"
+	"time"
+
 	"github.com/emicklei/go-restful"
+	"github.com/opensds/multi-cloud/api/pkg/common"
+	. "github.com/opensds/multi-cloud/api/pkg/s3/datatype"
+	s3error "github.com/opensds/multi-cloud/s3/error"
+	pb "github.com/opensds/multi-cloud/s3/proto"
 	log "github.com/sirupsen/logrus"
 	/*c "github.com/opensds/multi-cloud/api/pkg/context"
```
sfzeng: Same here.

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+			// Set headers on the first write.
+			// Set standard object headers.
+			SetObjectHeaders(response, object, hrange)
+
+			// Set any additional requested response headers.
+			setGetRespHeaders(response.ResponseWriter, request.Request.URL.Query())
+			dataWritten = true
+		}
+		n, err := response.Write(p)
+		if n > 0 {
```
sfzeng: What need to do if  n > 0?

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+
+		}
+		return n, err
+	})
+
+	s3err := int32(0)
```
sfzeng: Should the initial value of s3err be ErrNoErr?

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+		}
+		_, err = writer.Write(rsp.Data)
+		if err != nil {
+			log.Errorln("failed to write data to client. err:", err)
+			break
+		}
+		left -= int64(len(rsp.Data))
+	}
+
+	if !dataWritten {
```
sfzeng: dataWritten will always be 'true' because it was set on line 136.

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+
+		if rsp.ErrorCode != int32(s3error.ErrNoErr) {
+			s3err = rsp.ErrorCode
+			log.Errorf("received s3 service error, error code:", rsp.ErrorCode)
+			break
+		}
+		if len(rsp.Data) == 0 {
+			continue
+		}
+		_, err = writer.Write(rsp.Data)
```
sfzeng: Sorry, why don't we directly call 'response.Write(rsp.Data)' and set headers accordingly, instead of using funcToWriter here?

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+	if !dataWritten {
+		if s3err == int32(s3error.ErrNoErr) {
+			writer.Write(nil)
+		} else {
+			WriteErrorResponse(response, request, s3error.S3ErrorCode(s3err))
+			return
+		}
+	}
+
+	log.Info("PUT object successfully.")
```
sfzeng: It's GET instead of PUT here.

api/pkg/s3/objectget.go
```diff
@@ -24,65 +33,152 @@ import (
...
+			log.Errorln("recv err", err)
+			break
+		}
+		if err == io.EOF {
+			eof = true
+		}
+
+		if rsp.ErrorCode != int32(s3error.ErrNoErr) {
+			s3err = rsp.ErrorCode
+			log.Errorf("received s3 service error, error code:", rsp.ErrorCode)
```
sfzeng: Use log.Errorln or add the format string, like %d or %v here, please.

s3/pkg/service/object.go
```diff
@@ -256,9 +258,91 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+	}
+
+	eof := false
+	left := object.Size
+	buf := make([]byte, ChunkSize)
+	for !eof && left > 0 {
+		n, err := reader.Read(buf)
+		if err != nil && err != io.EOF {
+			log.Errorln("failed to read, err:", err)
+			break
```
sfzeng: Do we need to return this error to the api service in some way?

s3/pkg/service/object.go
```diff
@@ -256,9 +258,91 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+		}
+		if err == io.EOF {
+			log.Debugln("finished read")
+			eof = true
```
sfzeng: Same here.

s3/pkg/service/object.go
```diff
@@ -256,9 +258,91 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+		}
+		if n == 0 {
+			continue
+		}
+
+		err = stream.Send(&pb.GetObjectResponse{ErrorCode:0, Data:buf[0:n]})
```
sfzeng: Is it better to use  ErrNoErr instead of 0?

s3/pkg/service/object.go
```diff
@@ -256,9 +258,91 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+		n, err := reader.Read(buf)
+		if err != nil && err != io.EOF {
+			log.Errorln("failed to read, err:", err)
+			break
+		}
+		if err == io.EOF {
+			log.Debugln("finished read")
+			eof = true
+		}
+		if n == 0 {
```
sfzeng: Does this mean some errors happen?

s3/proto/s3.proto
```diff
@@ -378,7 +389,9 @@ message DeleteObjectInput {
 message GetObjectInput {
     string Bucket = 1;
     string Key = 2;
-    string Versionid = 3;
+    string VersionId = 3;
+	int64  Offset = 4;
```
sfzeng: Format it please.

api/pkg/s3/objectget.go
sunfch: A non-zero number of bytes may be returned with error io.eof.

api/pkg/s3/objectget.go
sunfch: thanks, I'll fix it

api/pkg/s3/api-headers.go
sunfch: ok, I'll fix it

api/pkg/s3/objectget.go
sunfch: thanks, I'll remove it

api/pkg/s3/objectget.go
sunfch: thanks, I'll remove it.  because I have code to implement other function at first.  now the function is not necessary

api/pkg/s3/objectget.go
sunfch: thanks, I'll fix it

api/pkg/s3/objectget.go
sunfch: if file size is zero, dataWritten is false

api/pkg/s3/objectget.go
sunfch: thanks,  I'll fix it

s3/pkg/service/object.go
sunfch: A zero number of bytes may be returned with error io.eof.

s3/proto/s3.proto
sunfch: ok, I'll fix it

s3/pkg/service/object.go
sunfch: ok, I'll fix it

api/pkg/s3/objectget.go
sunfch: because we only want to write some http header when write data

api/pkg/s3/objectget.go
sunfch: thanks, I'll fix it

s3/pkg/service/object.go
sunfch: I dont sure.  there is a situation that api service have sent some data to end user, and return the error to api service, it do nothing except print it

s3/pkg/service/object.go
sunfch: I dont sure.  there is a situation that api service have sent some data to end user, and return the error to api service, it do nothing except print it

s3/pkg/service/object.go
sfzeng: Then how to let the end user know there is something wrong when getting the object? Is it necessary?

api/pkg/s3/objectget.go
sfzeng: Does it seem the error of io.eof is not returned anywhere?

sunfch: 1. if all data have been sent to end user, we can ignore the error io.eof
2. if partial data have been sent,  the content-length is not equal to sent data length, and the request is  failed

s3/pkg/service/object.go
sunfch: because http header status code in response could has been sent to end user,  we can not send error info to user, and the request is failed because of partial download

api/pkg/s3/objectget.go
sunfch: I fix it to call writer directly

hopkings2008:
LGTM

hopkings2008:
LGTM

s3/pkg/service/object.go
kumarashit: a) can we put comment
b) Should we log  that zero bytes were returned. End of stream

s3/pkg/service/object.go
```diff
@@ -256,9 +258,91 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+	// if this object has only one part
+	backend, err := getBackend(ctx, s.backendClient, bucket)
+	if err != nil {
+		log.Errorln("unable to get backend. err:", err)
+		return err
+	}
+	sd, err := driver.CreateStorageDriver(backend.Type, backend)
+	if err != nil {
+		log.Errorln("failed to create storage driver. err:", err)
+		return nil
```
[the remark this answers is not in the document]
sunfch: Thanks, it is my mistake, and I'll fix it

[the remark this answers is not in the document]
sunfch: Thanks for your reviewing
when reader return non nil and eof error, we directly break loop and return error via defer func.
I add a comment below statement "if n == 0 {" to show zero bytes read, and when grpc server function return, the steam should be closed

s3/pkg/service/object.go
sunfch: The request will failed for partial download, and end user will receive a partial download error.  It is enough to print error info

s3/pkg/service/object.go
sunfch: Thanks for your reviewing, the comment and log info have been added

sunfch:
@kumarashit  
Thanks for your reviewing
when reader return non nil and eof error, we directly break loop and return error via defer func.
I add a comment below statement "if n == 0 {" to show zero bytes read, and when grpc server function return, the steam should be closed

PR #724, opened 2019-11-13
add multipart upload functions in yig storage backend

hopkings2008:
this pr contains the features of multipart upload and related tests in yig storage backend. They are ported from yig with some changes.

this pr contains the fix for https://github.com/opensds/multi-cloud/issues/714

**Release note**:

```release-note
```


s3/pkg/datastore/yig/conf/yig.toml
```diff
@@ -4,6 +4,7 @@ log_level=20
 
 [endpoint]
 url="default"
+machine_id=1
```
sfzeng: Sorry, machine_id will be used for what, and why it is set to be 1?

sfzeng: And why url is set to be 'default'? Does it need to be a traditional URL format string?

s3/pkg/datastore/yig/config/enum.go
```diff
@@ -40,7 +40,8 @@ func ReadConfigs(dir string, funcConfigParse FuncConfigParse) error {
 		viper.SetConfigFile(path)
 		err = viper.ReadInConfig()
 		if err != nil {
-			return err
+			// skip the config file which is failed to parse and continue the next one.
+			return nil
```
sfzeng: What if read config file failed? Anything will be used as default ?

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}()
+
+	_, err = tx.Exec(sqlText, partInfo.UploadId, partInfo.PartNum, partInfo.ObjectId, partInfo.Location, partInfo.Pool, partInfo.Offset, partInfo.Size, partInfo.Etag, partInfo.Flag)
+	if err != nil {
+		log.Errorf("failed to save partInfo(%v), err: %v", partInfo, err)
+		return err
+	}
+	// must return err instead of nil, because commit may return error.
+	return err
```
sfzeng: The error of tx.Commit() cannot be returned.

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
```diff
@@ -0,0 +1,145 @@
...
+
+	tx, err = t.DB.Begin()
+	if err != nil {
+		log.Errorf("failed to complete parts for uploadId(%d), it was fail to create transaction, err: %v", uploadId, err)
+		return err
+	}
+
+	defer func() {
+		if err == nil {
+			if err = tx.Commit(); err != nil {
```
sfzeng: Same here.

s3/pkg/datastore/yig/storage/multipart.go
```diff
@@ -1,53 +1,242 @@
...
+ */
+
 func (yig *YigStorage) InitMultipartUpload(ctx context.Context, object *pb.Object) (*pb.MultipartUpload, error) {
-	return nil, errors.New("not implemented.")
+	uploadId := uploadId2Str(yig.idGen.GetId())
+	mu := &pb.MultipartUpload{
+		Bucket:   object.BucketName,
+		Key:      object.ObjectKey,
+		UploadId: uploadId,
+		ObjectId: uploadId,
```
sfzeng: Why use uploadId for ObjectId?

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
```diff
@@ -0,0 +1,145 @@
...
+	stmt, err = tx.Prepare(sqlText)
+	if err != nil {
+		log.Errorf("failed to complete uploadId(%d), it was fail to prepare sql, err: %v", uploadId, err)
+		return err
+	}
+
+	defer stmt.Close()
+
+	for _, part := range parts {
+		_, err := stmt.Exec(part.Offset, part.Flag, uploadId, part.PartNum)
```
sfzeng: Why is responsible to set the Flag?

s3/pkg/datastore/yig/storage/multipart.go
hopkings2008: because the final object consists of more than one parts, and when complete all the parts, yig backend doesn't really perform combine operation to reduce the pressure of io on the ceph. So, we use uploadId to track all the parts which belongs to the object.

s3/pkg/datastore/yig/conf/yig.toml
hopkings2008: @sfzeng machine_id is used for machine id section of snowflake algo. and the "default" url is used for testing.

s3/pkg/datastore/yig/config/enum.go
hopkings2008: if we put some other file which is not toml file, viper ReadInConfig will return an error. In this case, we should skip this kind of file and continue to parse other files till we find our needed config files.

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
hopkings2008: @sfzeng , when error happens if we call tx.Commit(), that error will be returned to the caller, and  it will not be lost. the scope of err variable is in that whole function.

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
hopkings2008: the same as above comments.

s3/pkg/datastore/yig/config/enum.go
sfzeng: I mean ... if there is no valid config files, then what will happen?

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
sfzeng: Hi @hopkings2008 I tried the following program and got "test error!", please check:
package main

import (
        "fmt"
        "errors"
)

func test() error {
        var err error
        defer func() {
                err = errors.New("defer error!")
        }()

        err = errors.New("test error!")
        return err
}

func main() {
        err := test()
        fmt.Printf("err=%v\n", err)
}

hopkings2008: thanks very much @sfzeng for clarifying the error made by my original logic to return error. i have changed the logic of returning error by using named return value. pls help to check it again, thanks a lot.

s3/pkg/datastore/yig/config/enum.go
hopkings2008: the parse will fail, and we cannot get the configs needed by yig, so, yig will not create valid drivers and when request comes in, yig will return error when handle it.

s3/pkg/datastore/yig/meta/db/tidb/multiparts.go
hopkings2008: i have changed this process logic as above, pls help to check it again, thanks

PR #728, opened 2019-11-19
Support api of COPY object

### LINKED ISSUE #727: Support api of COPY object

> kind feature

**What happened**:

Support api of COPY object, include only updates of object attribute

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:
The requirement of the PR is from aws s3 api,  https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

The PR provide a api that is to create a copy of an object that is already stored in S3. The size of object to copy  must be less 5GB, because multipart of Copy Object is not implemented yet.

The implemented features are as follows:
1.  request with conditions.  x-amz-copy-source-if-match, x-amz-copy-source-if-none-match, .... 
2.  use "x-amz-metadata-directive" to replace with metadata provided in the request when source object is equal to target object.
3. copy object from a backend to another backend, or copy object in the same backend.

The most common use case is the following example.  It will create a copy of object "test.txt" in the same bucket
curl -H "Content-type: application/xml" -H "X-Amz-Copy-Source: /bucket001/test.txt" -X PUT http://127.0.0.1:8089/v1/s3/bucket001/test.txt2

Not implemented:
1. Access Permissions, ACL  ---  ACLs and access permissions are not involved in the PR when copy object
2. Versioning. the same as above
3. SSE. the same as above

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #727

**Release note**:

```release-note
```


api/pkg/s3/objectcopy.go
```diff
@@ -15,15 +15,204 @@
...
+
+	// Save sourceBucket and sourceObject extracted from url Path.
+	var err error
+	var sourceBucketName, sourceObjectName, sourceVersion string
+	if len(splits) == 2 {
+		sourceBucketName = splits[0]
+		sourceObjectName = splits[1]
+	}
+	// If source object is empty, reply back error.
+	if sourceObjectName == "" {
```
sfzeng: Do we need to check if sourceBucketName equals to "" too?

api/pkg/s3/objectcopy.go
```diff
@@ -15,15 +15,204 @@
...
+	}
+
+	/// maximum Upload size for object in a single CopyObject operation.
+	if isMaxObjectSize(sourceObject.Size) {
+		WriteErrorResponseWithResource(response, request, ErrEntityTooLarge, copySource)
+		return
+	}
+
+	// Note that sourceObject and targetObject are pointers
+	targetObject := &pb.Object{}
```
sfzeng: It seems targetObject is not used after it is defined.

api/pkg/s3/api-response.go
```diff
@@ -21,9 +21,11 @@ import (
 	"net/http"
```
rhsakarpos: Can you please point me to the Requirements document for the Copy Object  feature ? 
1. Specifically, is this only for AWS ?
2. if yes, how different is it from this https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

Another question, what is our behavior, during copy, w.r.t the following
SSE
Versioning
CORS
ACL
Storage class

hopkings2008:
LGTM

api/pkg/s3/objectcopy.go
sunfch: I think it is necessary, and how about your suggestion?

api/pkg/s3/objectcopy.go
sunfch: It seems that you are right,  I will delete it in the next commit, thanks

api/pkg/s3/api-response.go
sunfch: @rhsakarpos , Thanks for your reviewing.  The Requirement for the Copy Object is from AWS.  The PR implement partial features.  It include Copy Object between distinct backends and adding new attribute  and checking pre conditions in http header.  The PR dont support SSE, versioning, CORS, ACL, Storage class.

sfzeng: Hi @sunfch if the source object already set SSE, Versioning, etc, will those settings be copied, means set to new object too?

api/pkg/s3/objectcopy.go
sfzeng: I agree with you, if sourceBucketName equals to "", that's invalid. How do you think?

s3/pkg/service/object.go
```diff
@@ -363,6 +363,130 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
+	}
+	ctx = context.WithValue(ctx, dscommon.CONTEXT_KEY_SIZE, srcObject.Size)
+	ctx = context.WithValue(ctx, dscommon.CONTEXT_KEY_MD5, srcObject.Etag)
+	res, err := targetSd.Put(ctx, limitedDataReader, targetObject)
+	if err != nil {
+		log.Errorln("failed to put data. err:", err)
+		return err
+	}
+	if res.Written < srcObject.Size {
+		// TODO: delete incomplete object at backend
```
sfzeng: Hi @sunfch  Could you please add a issue to trace those TODO things?

sunfch: I have add a issue #740 to trace it

api/pkg/s3/api-response.go
sunfch: @sfzeng , thanks for your review,  SSE and Versioning are not include in the PR now

s3/pkg/meta/types/object.go
```diff
@@ -223,3 +223,12 @@ func (o *Object) GetCreateSql() (string, []interface{}) {
 
 	return sql, args
 }
+
+func (o *Object) GetUpdateMetaSql() (string, []interface{}) {
+	version := math.MaxUint64 - uint64(o.LastModified)
+	attrs, _ := json.Marshal(o.CustomAttributes)
+	sql := "update objects set customattributes =? where bucketname=? and name=? and version=?"
```
sfzeng: From objectcopy.go in api, it seems user can update customattritubes, storage class, and content type, but here, only customattritubes will updated. Is that a bug?

s3/pkg/service/object.go
```diff
@@ -363,6 +363,130 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
+	if err != nil {
+		log.Errorln("failed to get backend client with err:", err)
+		return err
+	}
+	srcSd, err := driver.CreateStorageDriver(srcBackend.Type, srcBackend)
+	if err != nil {
+		log.Errorln("failed to create storage. err:", err)
+		return err
+	}
+	reader, err := srcSd.Get(ctx, srcObject.Object, 0, srcObject.Size)
```
sfzeng: IMHO, it's better to get target bucket and  target backend, and create target storage driver before get data from source backend.

s3/pkg/meta/types/object.go
sunfch: @sfzeng , Thanks for your reviewing, this is a bug. I"ll fix it in next commit

s3/pkg/service/object.go
sunfch: thanks for your reviewing, I will do it as you proposed in next commit

api/pkg/s3/objectcopy.go
sunfch: Thanks for your reviewing, I will add sourceBucketName condition check to it

s3/pkg/meta/types/object.go
sunfch: acl and content type are updated in latest commit.  storage class have not been updated, because there is no storage class in db schema. I add a issue to trace it

api/pkg/s3/objectcopy.go
```diff
@@ -15,15 +15,194 @@
...
+		}
+
+		log.Info("COPY object successfully.")
+		// write success response.
+		WriteSuccessResponse(response, encodedSuccessResponse)
+		return
+	}
+
+	/// maximum Upload size for object in a single CopyObject operation.
+	if isMaxObjectSize(sourceObject.Size) {
```
kumarashit: Should this check be done at the start? If we are not allowing bigger size object, we should discard upfront than coming all down here and then error out

api/pkg/s3/objectcopy.go
sunfch: @kumarashit  Thanks,  I add some lines to return error when the parts of len(splits) is not equal 2

api/pkg/s3/objectcopy.go
```diff
@@ -15,15 +15,194 @@
...
+
+	splits = strings.SplitN(sourceObjectName, "?", 2)
+	if len(splits) == 2 {
+		sourceObjectName = splits[0]
+		if !strings.HasPrefix(splits[1], "versionId=") {
+			WriteErrorResponse(response, request, ErrInvalidCopySource)
+			return
+		}
+		sourceVersion = strings.TrimPrefix(splits[1], "versionId=")
+	}
```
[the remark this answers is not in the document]
sunfch: Thanks for your reviewing, the checking for sourceBucketName have been added. I have add a few lines code to returned error when len(splits) != 2  in next commit

api/pkg/s3/objectcopy.go
```diff
@@ -15,15 +15,194 @@
...
+			response.ResponseWriter.Header()["ETag"] = []string{"\"" + result.Md5 + "\""}
+		}
+		if sourceVersion != "" {
+			response.AddHeader("x-amz-copy-source-version-id", sourceVersion)
+		}
+		if result.VersionId != "" {
+			response.AddHeader("x-amz-version-id", result.VersionId)
+		}
+
+		log.Info("COPY object successfully.")
```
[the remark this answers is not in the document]
sunfch: thanks for your reviewing,  I will fix it as your suggestion in next commit.

api/pkg/s3/objectcopy.go
sunfch: @kumarashit, Thanks for your reviewing,  we can update object meta although we are not allowing bigger size object.  if we place this check code at start of the api, we cannot update meta for bigger size object.

api/pkg/s3/objectcopy.go
```diff
@@ -15,15 +15,194 @@
...
+		return
+	}
+
+	//TODO: In a versioning-enabled bucket, you cannot change the storage class of a specific version of an object. When you copy it, Amazon S3 gives it a new version ID.
+	storageClassFromHeader, err := getStorageClassFromHeader(request)
+	if err != nil {
+		WriteErrorResponse(response, request, err)
+		return
+	}
+	if storageClassFromHeader == types.ObjectStorageClassGlacier || storageClassFromHeader == types.ObjectStorageClassDeepArchive {
```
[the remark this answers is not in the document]
sunfch: Thanks, I have add a new issue to trace it, #745

api/pkg/common/common.go
```diff
@@ -30,7 +30,7 @@ const (
 	MaxPaginationLimit      = 1000
 	DefaultPaginationLimit  = MaxPaginationLimit
 	DefaultPaginationOffset = 0
-	MaxObjectSize           = 5 * 1024 * 1024 // 5GB
+	MaxObjectSize           = 5 * 1024 * 1024 * 1024 // 5GB
```
kumarashit: Just confirming..1024 is the standard right? no 1000. please confirm GB or GiB

sunfch: @kumarashit ,  Thanks for your reviewingm,  I created a 5*1024*1024*1024 size file and put it to aws,  and aws response with ok, so it should be 1024

Issue #638, opened 2019-10-25
GET bucket SSE-YIG


rhsakarpos:

rhsakarpos:
@sfzeng please reopen in case needed, marking this closed

PR #729, opened 2019-11-19
SSE support

### LINKED ISSUE #634: PUT bucket SSE

rhsakarpos:
**Support SSE in OpenSDS. This PR adds support for the server generated random key encryption. The SSE-C and SSE-KMS are not targeted for DAITO**:

**Fixes the SSE related issue for DAITO**: fixes #634 #638 

**Release note**:

```The way this works is to encrypt the object byte stream in OSDS and upload to cloud bucket. Similarly, decrypt the byte stream on GET object (this is not done yet, waiting on GET to be implemented). The encryption settings on the cloud bucket are NOT altered. We use AES-256 with a random key, to encrypt the data before a PUT object happens

Further, in the current code, I am yet to get enough details on the following topics, will integrate my code with these once I have the information and complete the production testing with docker

1. AK-SK
2. Policies
3. Tenant ID

Tested the following on my local environment (without docker), using postman
1. PUT SSE option on bucket
2. GET SSE option on listing buckets
3. Encrypt data byte[ ] on PUT object
```


rhsakarpos:
@skdwriting @wisererik @sfzeng @nguptaopensds @kumarashit @anvithks @sunfch  please review, thanks

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,172 @@
...
+		s.s3Client.UpdateBucket(ctx, bucket.BucketMeta)
+
+		createLifecycleConf := model.LifecycleConfiguration{}
+		err := xml.Unmarshal(body, &createLifecycleConf)
+		if err != nil {
+			response.WriteError(http.StatusInternalServerError, err)
+			return
+		} else {
+			dupIdCheck := make(map[string]interface{})
+			s3RulePtrArr := make([]*s3.LifecycleRule, 0)
```
nguptaopensds: Lifecycle Code is there Prakash. You can review the file again, remove that code

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	body := ReadBody(request)
+	log.Infof("MD5 sum for body is %x", md5.Sum(body))
+
+	if body != nil {
+		sseConf := model.SSEConfiguration{}
+		errSSE := xml.Unmarshal(body, &sseConf)
+		if errSSE != nil {
+			response.WriteError(http.StatusInternalServerError, err)
```
nguptaopensds: Prakash we have changed the error response in development branch, based on yig changes. Now we use WriteErrorResponse instead of response.WriteError. Please follow updated development branch to code further

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,174 @@
...
+
+	"github.com/emicklei/go-restful"
+	"github.com/opensds/multi-cloud/api/pkg/common"
+	. "github.com/opensds/multi-cloud/api/pkg/utils/constants"
+	"github.com/opensds/multi-cloud/s3/pkg/model"
+	s3 "github.com/opensds/multi-cloud/s3/proto"
+	log "github.com/sirupsen/logrus"
+)
+
+// Map from storage calss to tier
```
skdwriting: class ;)

api/pkg/s3/bucketsseput.go
skdwriting: > Lifecycle Code is there Prakash. You can review the file again, remove that code

lifecycle is not supported?

api/pkg/s3/objectput.go
```diff
@@ -146,7 +148,14 @@ func (s *APIService) ObjectPut(request *restful.Request, response *restful.Respo
 			log.Debugln("finished read")
 			eof = true
 		}
-		err = stream.Send(&s3.PutObjectRequest{Data: buf[:n]})
+		// encrypt if needed
+		if bucketMeta.ServerSideEncryption.SseType == "SSE"{
+			key := bucketMeta.ServerSideEncryption.EncryptionKey
+			_, encbuf := utils.EncryptWithAES256RandomKey(buf, key)
```
skdwriting: always with AES256RandomKey? No encryption config?

api/pkg/s3/routbucketdelete.go
```diff
@@ -38,7 +38,10 @@ func (s *APIService) RouteBucketDelete(request *restful.Request, response *restf
 	} else if IsQuery(request, "lifecycle") {
 		s.BucketLifecycleDelete(request, response)
 
-	} else {
+	}else if IsQuery(request, "DefaultEncryption") {
+		s.BucketSSEDelete(request, response)
+
+	}else {
 		s.BucketDelete(request, response)
```
skdwriting: when will it hit this?

api/pkg/utils/encryptionutils.go
```diff
@@ -0,0 +1,77 @@
+package utils
```
skdwriting: license?

api/pkg/s3/bucketsseput.go
nguptaopensds: No this file is bucketSSEput, there is some lifecycle related code in this , I just asked to remove that

rhsakarpos: okay, can i remove it after the RC ?

api/pkg/s3/bucketsseput.go
rhsakarpos: okay, will do during testing phase, is that okay ?

api/pkg/s3/bucketsseput.go
rhsakarpos: oops, can i fix after RC, cannot push to github right now

api/pkg/s3/objectput.go
rhsakarpos: this is the server managed key variant, so server picks a random key and uses AES-256, same as AWS model

api/pkg/s3/routbucketdelete.go
rhsakarpos: on DELETE of encryption on bucket. not supported for DAITO from UI

api/pkg/utils/encryptionutils.go
rhsakarpos: license ?

rhsakarpos: got it..updated

kumarashit:
@rhsakarpos Please update the micro and pb files conflict

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,174 @@
...
+		errSSE := xml.Unmarshal(body, &sseConf)
+		if errSSE != nil {
+			response.WriteError(http.StatusInternalServerError, err)
+			return
+		}
+
+		s3SSE  := &s3.ServerSideEncryption{
+			SseType:              "",
+			EncryptionKey:        nil,
+			InitilizationVector:  nil,
```
hopkings2008: @rhsakarpos it seems that the InitilizationVector is not used, but for encryption algo such as aes, initilization is needed.

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -535,3 +554,78 @@ func (t *TidbClient) ListBucketLifecycle(ctx context.Context) (buckets []*Bucket
...
+		return handleDBError(err)
+	}
+
+	return nil
+}
+
+func (t *TidbClient) CreateBucketSSE(ctx context.Context, bucketName string, sseType string, sseserverkey string) error {
+	log.Infof("create bucket[%s] SSE info[%s] into tidb ...\n", bucketName, sseType)
+
+	sql := "insert into bucket_sseopts(bucketname, sse, sseserverkey) values(?,?,?);"
```
hopkings2008: @rhsakarpos one question here, why the sse info is saved into a new table bucket_sseopts instead of the bucket table? if the sse info can be saved into bucket table, the bucket_sseopts table is not needed then.
And for Object put operation, we need to save the sse info into object table too, because put object operation can have sse header and it's priority  is higher than the one in bucket.
https://docs.aws.amazon.com/AmazonS3/latest/dev/bucket-encryption.html 
'If your PUT request headers don't include encryption information, Amazon S3 uses the bucket’s default encryption settings to encrypt the objects.

If your PUT request headers include encryption information, Amazon S3 uses the encryption information from the PUT request to encrypt objects before storing them in Amazon S3.'

api/pkg/s3/bucketsseput.go
rhsakarpos: hi @hopkings2008 not really, IV is not needed for AES 256. If we use a custom salt, then we need an IV, this encrypts just fine. Thanks.

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: @hopkings2008 yes, this is done to normalize the table design. Rather than storing all info in a single table, with so many columns, we can have a bucket table, and have other tables like bucket_sse and bucket_versioning etc. to store specifics related to these. Makes it easier to extend these features and not affect the basic bucket features

rhsakarpos: Also, for the header comment, you are right. This header is not supported in DAITO. Why ? I need to evaluate ALL cloud backends to see if they support this header (AWS does), then will add to OSDS. Thanks.

hopkings2008: > @hopkings2008 yes, this is done to normalize the table design. Rather than storing all info in a single table, with so many columns, we can have a bucket table, and have other tables like bucket_sse and bucket_versioning etc. to store specifics related to these. Makes it easier to extend these features and not affect the basic bucket features

@rhsakarpos sse info is specific to that bucket, and the bucket cannot have more than sse settings in the same time, right? so, for this sse case, i think it is not suitable to have one separate table for sse. and also, retrieve that information will need two queries. normalization is used to reduce the level of redundancy, since sse is not redundancy here, so the separate table is not needed. The same as bucket versioning.

hopkings2008: > Also, for the header comment, you are right. This header is not supported in DAITO. Why ? I need to evaluate ALL cloud backends to see if they support this header (AWS does), then will add to OSDS. Thanks.

sse header in object is defined by s3, users will access multi-cloud through s3 protocol, so multi-cloud needs to be compatible with s3 protocol, api layer must confirm this. if backend doesn't support sse, we can return an error to user.

api/pkg/s3/bucketsseput.go
hopkings2008: @rhsakarpos because the core of encryption is to make the output message after encrypition more randomly, and currently, by using the iv for block-based cipher, the output message after encrypted will be more randomly. So, if we want to make the encryption more secure, it is better to use iv.  Just my opinion :)

hopkings2008: because the key generated in bucket sse put api is used to encrypt all the objects, so , iv should be use, pls refer to the link 'https://security.stackexchange.com/questions/35210/encrypting-using-aes-256-do-i-need-iv', thanks

rhsakarpos: thanks @hopkings2008, will add post DAITO. Raised #750 to track this, assigned to me

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: @hopkings2008 , true. However, if the other cloud backend do NOT support this feature of a custom header, then do we still need the access header support ?

rhsakarpos: about the DB table, yes, one bucket has one SSE setting at a time. Duplication apart, there is a separation of concern and low/no impact on existing code, if we keep the schema separate for bucket information, version information, SSE information etc. If any feature adds or removes columns, constraints, primary keys etc. it does not affect the other features

api/pkg/s3/router.go
```diff
@@ -44,4 +44,8 @@ func RegisterRouter(ws *restful.WebService) {
 	ws.Route(ws.PUT("/{bucketName}/?lifecycle").To(handler.RouteBucketPut)).Doc("Create lifecycle configuration for the bucket")
 	ws.Route(ws.GET("/{bucketName}/?lifecycle").To(handler.RouteBucketGet)).Doc("Get lifecycle configuration from the bucket")
 	ws.Route(ws.DELETE("/{bucketName}/?lifecycle").To(handler.RouteBucketDelete)).Doc("Delete lifecycle configuration from the bucket")
+	ws.Route(ws.DELETE("/{bucketName}/?DefaultEncryption").To(handler.RouteBucketDelete)).Doc("Delete lifecycle configuration from the bucket")
+
+	// router for SSE
+	ws.Route(ws.PUT("/{bucketName}/?DefaultEncryption").To(handler.RouteBucketPut)).Doc("Set default encryption on bucket")
```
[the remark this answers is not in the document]
rhsakarpos: yes, will check on how much the other cloud backend are compliant to this specification in entirety and update

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,174 @@
...
+			XXX_sizecache:        0,
+		}
+		if sseConf.SSE.Enabled=="true"{
+			s3SSE.SseType = "SSE"
+			s3SSE.EncryptionKey,_ = utils.GetRandom32BitKey()
+		}
+
+		bucket.BucketMeta.ServerSideEncryption = s3SSE
+
+		s.s3Client.UpdateBucket(ctx, bucket.BucketMeta)
```
sfzeng: Hi @rhsakarpos , please think about the scenario that user create a bucket without SSE first, and set SSE later, it seems for this scenario, SSE can not be set successfully, cause nothing will change for UpdateBucket as there is no record in bucket_sseopts.

api/pkg/s3/bucketsseput.go
skdwriting: Please do it as RC1 fix

api/pkg/s3/objectput.go
skdwriting: we will discuss. I meant whether we consider any other encryption or not.

api/pkg/s3/routbucketdelete.go
skdwriting: Why ? Also why only Delete from Dashboard special?

api/pkg/s3/bucketsseput.go
skdwriting: Raised an issue #759 to track

s3/pkg/meta/db/drivers/tidb/bucket.go
hopkings2008: one more thing that, if we add one table each time for a new bucket attribute, we will query much time when handle a request, this will introduce a big pressure for the backend db. Also, so many tables will introduce too process logics in a handler, this may introduce problems for maintenance.

kumarashit:
@rhsakarpos Can you please check to resolve conflicts and then we will check how to handle the CI issue by issuing /retest

rhsakarpos:
closing this due to some CI issues and merge conflicts, opened #758

rhsakarpos:
closing this due to some CI issues and merge conflicts, opened #758

PR #730, opened 2019-11-19
Implement delete interface in yig storage backend

hopkings2008:
this pr implements the delete interface in yig storage backend. And to reduce the io pressure for ceph, yig will start a gc to remove the objects from ceph. Each call to delete interface provided by yig storage backend, it will add a record into gc table of yig. Then, gc will read that table and clear the objects from ceph.
Also, the related tests are added and all the test cases of yig storage backend are passed in my local env.

**Release note**:

```release-note
```


s3/pkg/datastore/yig/conf/yig.sql
```diff
@@ -43,19 +43,28 @@ DROP TABLE IF EXISTS `multiparts`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `multiparts` (
```
sfzeng: I suggested to create a seperate database from s3.

s3/pkg/datastore/yig/conf/yig.sql
```diff
@@ -43,19 +43,28 @@ DROP TABLE IF EXISTS `multiparts`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
```
sfzeng: Is this tables still in the database used by s3?  IMHO, it's better to have a spereate database for yig backend.

hopkings2008: @sfzeng these tables are only used by yig backend. and yig backend has a separate db which is insulated from the one used by s3.

sfzeng: OK, thanks！

s3/pkg/datastore/yig/factory.go
```diff
@@ -56,18 +68,17 @@ func (ydf *YigDriverFactory) Init() error {
 	}
 
 	// read common config settings
-	cc, err := config.ReadCommonConfig("/etc/yig")
+	/*cc, err := config.ReadCommonConfig("/etc/yig")
```
sfzeng: Remove unused codes please.

s3/pkg/datastore/yig/storage/object.go
```diff
@@ -233,8 +232,58 @@ func (yig *YigStorage) Get(ctx context.Context, object *pb.Object, start int64,
...
+* Below is the process logic:
+* 1. check whether objectId is multipart uploaded, if so
+* retrieve all the object ids from multiparts and put them into gc.
+* or else, put it into gc.
+*
+ */
 func (yig *YigStorage) Delete(ctx context.Context, object *pb.DeleteObjectInput) error {
-	return errors.New("not implemented.")
+	// check multipart uploaded object.
+	if object.StorageMeta == "" {
```
sfzeng: Could you please add comment to explain that's the scenario when object.StorageMeta equals ""?

s3/pkg/datastore/yig/storage/gc.go
```diff
@@ -0,0 +1,281 @@
...
+	ctx        context.Context
+	cancelFunc context.CancelFunc
+	yig        *YigStorage
+	loopTime   int64
+	wg         sync.WaitGroup
+}
+
+func (gm *GcMgr) Start() {
+	// query the current available cpus
+	threadNum := runtime.GOMAXPROCS(0)
```
sfzeng: How many routines we will start? Will that use too much cpu resource?

s3/pkg/datastore/yig/factory.go
hopkings2008: @sfzeng thanks for your suggestion, i have removed these unused codes, pls have a check again, thanks.

s3/pkg/datastore/yig/storage/object.go
hopkings2008: @sfzeng thanks for your suggestion, i have added comments for such check, pls review it again, thanks.

s3/pkg/datastore/yig/storage/gc.go
hopkings2008: @sfzeng thanks for your questions here, if multi-cloud process is running in the docker container, this call will return all the available cpus despite there is a cpu restriction for this container, and https://github.com/uber-go/automaxprocs is used to fix this problem. Because that 3rd lib will introduce so many files in this pr, could i open an issue to track this problem and fix this problem in a separate pr? thanks a lot.

sfzeng: OK, please open an issue, and fix in later PR, thanks !

hopkings2008: @sfzeng thanks a lot, i have created issue https://github.com/opensds/multi-cloud/issues/741 to track this problem you found. i will submit a separate pr for this issue soon.

s3/pkg/datastore/yig/conf/yig.sql
```diff
@@ -43,19 +43,28 @@ DROP TABLE IF EXISTS `multiparts`;
...
+  `update_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
   PRIMARY KEY (`upload_id`, `part_num`),
   UNIQUE KEY `rowkey` (`upload_id`,`part_num`,`object_id`, `flag`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+DROP TABLE IF EXISTS `gc` 
+CREATE TABLE `gc` (
+	`id` bigint(20) PRIMARY KEY AUTO_INCREMENT,
+	`location` varchar(255) NOT NULL DEFAULT '',
```
kumarashit: What is the location?

hopkings2008: @kumarashit thanks for your review, the location is used by yig storage backend and it is the ceph cluster identifier.

Issue #713, opened 2019-11-09
Support versioning of bucket


sfzeng:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 

> /kind feature

**Anything else we need to know?**:
To support versioning, we need to think:
1. APIs for PUT/GET/DISABLE/ENABLE versioning.
2. What will happen when we PUT an object if versioning is enabled/suspended.
3. What will happen when we get an object if versioning is enabled/suspended.
3. What will happen when we list objects of a bucket versioning is enabled/suspended.
4. What will happen when we get an object if versioning is enabled/suspended.
5. What will happen when we delete an object if versioning is enabled/suspended.

**Environment**:
- Gelato(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nguptaopensds:
PUT bucket versioning on all cloud backends issues : #608 #609 #610 #611 #612 
DELETE bucket versioning on all cloud backends issues : #613 #614 #615 #616 #617

nguptaopensds:
Issue #736 takes care of Versioning DELETE and DELETE marker support on object level

PR #739, opened 2019-11-21
PUT bucket Versioning API for multi-cloud

nguptaopensds:
This is the first PR for support versioning in multi-cloud, It has PUT bucket versioning API changes.

#713 
1. This has initial changes for PUT bucket versioning where the metadata is updated.
2. Next we will include the versioning logic to get and set next version of object.
Testing :
1. Tested PUT bucket versioning on my environment using POSTMAN and cross checked from MySQL workbench also

**Release note**:

```release-note
```


api/pkg/s3/bucketversioningput.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func (s *APIService) BucketVersioningPut(request *restful.Request, response *restful.Response) {
+	bucketName := request.PathParameter("bucketName")
+	log.Infof("received request for creating versioning of bucket: %s", bucketName)
+
+	ctx := common.InitCtxWithAuthInfo(request)
+	bucket, err := s.s3Client.GetBucket(ctx, &s3.Bucket{Name: bucketName})
+	if HandleS3Error(response, request, err, bucket.ErrorCode) != nil {
+		log.Errorf("get bucket[%s] lifecycle failed, err=%v, errCode=%d\n", bucketName, err, bucket.ErrorCode)
```
kumarashit: Correct the error message

api/pkg/s3/bucketversioningput.go
```diff
@@ -0,0 +1,77 @@
...
+		return
+	}
+
+	body := ReadBody(request)
+	log.Infof("MD5 sum for body is %x", md5.Sum(body))
```
kumarashit: If body is nil will MD5 work on that? should this be after the error handling?

api/pkg/s3/bucketversioningput.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if versionConf.Status == "Enabled"{
+		s3version.Status = "Enabled"
+	} else{
+		s3version.Status = "Disabled"
+	}
+
+	bucket.BucketMeta.Versioning = s3version
+
+	rsp, err := s.s3Client.CreateBucket(ctx, bucket.BucketMeta)
```
kumarashit: Should we call Update Bucket?

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -167,6 +175,15 @@ func (t *TidbClient) GetBuckets(ctx context.Context) (buckets []*Bucket, err err
 			err = handleDBError(err)
 			return
 		}
+
+		//get versioning for the bucket
+		versionOpts,versionErr := t.GetBucketVersioning(ctx, tmp.Name)
+		if versionErr != nil {
+			return
```
kumarashit: Log??

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -535,3 +552,66 @@ func (t *TidbClient) ListBucketLifecycle(ctx context.Context) (buckets []*Bucket
...
+
+	sql := "update bucket_versionopts set versionstatus=? where bucketname=?"
+	args := []interface{}{versionStatus, bucketName}
+
+	_, err := t.Client.Exec(sql, args...)
+	if err != nil {
+		return handleDBError(err)
+	}
+
+	return nil
```
kumarashit: Put more logs

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -535,3 +552,66 @@ func (t *TidbClient) ListBucketLifecycle(ctx context.Context) (buckets []*Bucket
...
+		return nil, ErrInternalError
+	}
+
+	var rows *sql.Rows
+	sqltext := "select versionstatus from bucket_versionopts where bucketname=?;"
+
+	rows, err = t.Client.Query(sqltext, bucketName)
+
+	if err == sql.ErrNoRows {
+		err = nil
```
kumarashit: Put logs

s3/pkg/service/service.go
```diff
@@ -395,12 +403,14 @@ func (s *s3Service) GetBucketVersioning(ctx context.Context, in *pb.BaseBucketRe
 	return nil
 }
 
-func (s *s3Service) PutBucketVersioning(ctx context.Context, in *pb.PutBucketVersioningRequest, out *pb.BaseResponse) error {
+/*func (s *s3Service) PutBucketVersioning(ctx context.Context, in *pb.PutBucketVersioningRequest, out *pb.BaseResponse) error {
```
kumarashit: Put a comment or TODO why we have commented this code

s3/pkg/conf/tidb.sql
```diff
@@ -239,3 +239,8 @@ CREATE TABLE IF NOT EXISTS `lifecycle` (
                        `status` varchar(255) DEFAULT NULL
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin;
 
+CREATE TABLE IF NOT EXISTS `bucket_versionopts` (
```
sfzeng: Hi It seems, only one status saved for versioning, so why we created a new table bucket_versionopts, but not used bucket table itself?

api/pkg/s3/bucketversioningput.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	s3version := &s3.BucketVersioning{
+		Status:               "",
+		XXX_NoUnkeyedLiteral: struct{}{},
+		XXX_unrecognized:     nil,
+		XXX_sizecache:        0,
+	}
+
+	if versionConf.Status == "Enabled"{
```
sfzeng: Could you please use const instead of string directly?

api/pkg/s3/bucketversioningput.go
```diff
@@ -0,0 +1,77 @@
...
+		s3version.Status = "Enabled"
+	} else{
+		s3version.Status = "Disabled"
+	}
+
+	bucket.BucketMeta.Versioning = s3version
+
+	rsp, err := s.s3Client.UpdateBucket(ctx, bucket.BucketMeta)
+	if HandleS3Error(response, request, err, rsp.ErrorCode) != nil {
```
sfzeng: It seems rsp.ErrorCode is not set by s3Client.UpdateBucket, so we cannot get the specific s3 error as supposed.

api/pkg/s3/listbuckets.go
```diff
@@ -38,7 +38,13 @@ func parseListBuckets(list *s3.ListBucketsResponse) []byte {
 	buckets := []model.Bucket{}
 	for _, value := range list.Buckets {
 		ctime := time.Unix(value.CreateTime, 0).Format(time.RFC3339)
-		bucket := model.Bucket{Name: value.Name, CreateTime: ctime, LocationConstraint: value.DefaultLocation}
+		versionOpts := model.VersioningConfiguration{}
+		if value.Versioning != nil{
+			if value.Versioning.Status == "Enabled"{
```
sfzeng: Better to use const constant than string("Enabled") here.

api/pkg/s3/listbuckets.go
```diff
@@ -38,7 +38,13 @@ func parseListBuckets(list *s3.ListBucketsResponse) []byte {
 	buckets := []model.Bucket{}
 	for _, value := range list.Buckets {
 		ctime := time.Unix(value.CreateTime, 0).Format(time.RFC3339)
-		bucket := model.Bucket{Name: value.Name, CreateTime: ctime, LocationConstraint: value.DefaultLocation}
+		versionOpts := model.VersioningConfiguration{}
```
sfzeng: Versioning is not get from database when list buckets, so it's invaluable to get versioning here.

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -110,6 +111,13 @@ func (t *TidbClient) GetBucket(ctx context.Context, bucketName string) (bucket *
 		err = handleDBError(err)
 		return
 	}
+	//get versioning for the bucket
+	versionOpts,versionErr := t.GetBucketVersioning(ctx, tmp.Name)
+	if versionErr != nil {
+		return
```
sfzeng: 1. versionErr will not be returned here.
2. Add an error log please.

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -110,6 +111,13 @@ func (t *TidbClient) GetBucket(ctx context.Context, bucketName string) (bucket *
...
+	}
+	tmp.Versioning = &pb.BucketVersioning{}
```
sfzeng: Only status needed, why use a new struct, BucketVersioning, here?

s3/pkg/service/service.go
```diff
@@ -302,6 +302,14 @@ func (s *s3Service) GetTierMap(ctx context.Context, in *pb.BaseRequest, out *pb.
 func (s *s3Service) UpdateBucket(ctx context.Context, in *pb.Bucket, out *pb.BaseResponse) error {
 	log.Info("UpdateBucket is called in s3 service.")
 
+	//update versioning if not nil
+	if in.Versioning != nil {
+		err := s.MetaStorage.Db.UpdateBucketVersioning(ctx, in.Name, in.Versioning.Status)
```
sfzeng: What if user create a bucket without versioning, and then put versioning to this bucket? In this scenario, nothing will happen by update, because you created a new table to save versioning and no record in this table.

s3/proto/s3.proto
```diff
@@ -273,7 +268,7 @@ message Bucket {
 	string id = 6;
 	bool deleted = 7;
 	ServerSideEncryption serverSideEncryption = 8;
-	string versioning = 9;
+	BucketVersioning versioning = 9;
```
sfzeng: Same question, only status need for versioning, why add a new struct here?

api/pkg/s3/bucketversioningput.go
nguptaopensds: Sure Will update that

api/pkg/s3/bucketversioningput.go
nguptaopensds: Will put error handling in Update bucket also

api/pkg/s3/listbuckets.go
nguptaopensds: Yes will update

s3/pkg/conf/tidb.sql
nguptaopensds: For example, if in future we want to include versionid and version details also, the new table can be of reference easily, hence created new one

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: Sure will add

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: Earlier versioning was just a boolean value, yes or no. but we need different details for example time stamp and versioning info, hence created another struct for that

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: will put

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: yes will put

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: yes will put

s3/pkg/service/service.go
nguptaopensds: When user will get the interface of createbucket on dashboard there will be a button to select versioning enabled or disabled by default, so always we will get one status value, hence it is handled like that

s3/pkg/service/service.go
nguptaopensds: done

s3/proto/s3.proto
nguptaopensds: addressed above

s3/pkg/conf/tidb.sql
sfzeng: But... versionid is for object not bucket, what we are talking about is bucket.

s3/pkg/meta/db/drivers/tidb/bucket.go
sfzeng: Time stamp and versioning info are for object, right? But what you created is only for bucket.

s3/pkg/service/service.go
sfzeng: Then in the scenario I mentioned above, which API is supposed to be used? No record in the new table you created while updating it, so nothing will happen.

api/pkg/s3/bucketversioningput.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		WriteErrorResponse(response, request, S3ErrorCode(ErrInvalidVersion))
+		return
+	}
+
+	s3version := &s3.BucketVersioning{
+		Status:               "",
+		XXX_NoUnkeyedLiteral: struct{}{},
+		XXX_unrecognized:     nil,
+		XXX_sizecache:        0,
```
sunfch: The XXX_ types are used by the Protobuf library to store unknown fields,  and we don't need to set them?

s3/error/s3errors.go
```diff
@@ -650,6 +651,11 @@ var ErrorCodeResponse = map[S3ErrorCode]S3ErrorStruct{
 		Description:    "The LC configuration specified in the request is invalid.",
 		HttpStatusCode: http.StatusBadRequest,
 	},
+	ErrInvalidVersion: {
+		AwsErrorCode:   "NoSuchBucketVersion",
+		Description:    "The Versioning configuration specified in the request is invalid.",
+		HttpStatusCode: http.StatusBadRequest,
+	},
```
sunfch: It seems that the "ErrInvalidVersioning" defined in the file is confused  with the "ErrInvalidVersion",   and the AwsErrorCode for "ErrInvalidVersion" is not in https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

s3/pkg/conf/tidb.sql
sunfch: Buckets can be in one of three states: unversioned (the default), versioning-enabled, or versioning-suspended, and I think    
it is better that we store the version states in table bukets

api/pkg/s3/bucketversioningput.go
nguptaopensds: I  guess without setting them also those will take default values

s3/error/s3errors.go
nguptaopensds: Let me check that

s3/pkg/conf/tidb.sql
nguptaopensds: Status is just one of the attributes in bucket version table, in future we may need to add/modify different attributes and each time to alter bucket table is not wise , hence to normalize the bucket table it's better we use different table for versioning query, say for versioning if someone wants to add or modify parameters, they can just modify version table, no changes on listbucket, getbucket modules. just change in versioning related APIs

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: Okay will update it after RC1, we need to think more over versioning struct, whether we need few more details to be added, for example AWS has "MFA delete" attribute also apart from "Status", so if we also need some attribute alike AWS , we can modify versioning struct

s3/pkg/service/service.go
nguptaopensds: Sure will check

s3/error/s3errors.go
nguptaopensds: Updated

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: Done

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: Done

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -160,13 +172,24 @@ func (t *TidbClient) GetBuckets(ctx context.Context) (buckets []*Bucket, err err
...
 			&updateTime)
 		if err != nil {
 			err = handleDBError(err)
 			return
 		}
+
+		//get versioning for the bucket
+		versionOpts, versionErr := t.GetBucketVersioning(ctx, tmp.Name)
+		if versionErr != nil {
+			return
```
kumarashit: Put a error log here

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: Done

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: done

s3/pkg/meta/db/drivers/tidb/bucket.go
sfzeng: Is MFA related to versioning? What kind of attribute may need to be added for versioning?

nguptaopensds: YEs MFA is related to versioning, and we can have delete Marker added as an attribute in near future

s3/pkg/service/service.go
sfzeng: Hi @nguptaopensds how about this?

PR #744, opened 2019-11-23
add listparts interface into driver.go of storage backend.

hopkings2008:
**What this PR does / why we need it**:
this pr adds interface of listing parts for multipart upload in storage backend driver.
currently, yig and huawei storage backend implements the ListParts interface in this pr, and aws will implement it later.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #746, opened 2019-11-25
remove the old object from backend when put the object with the same key in same bucket more than once

### LINKED ISSUE #765: No database created for yig storage backend in docker-compose.yml
This is a bug
>
> /kind bug
> /kind feature

**What happened**:
No database created for yig storage backend in docker-compose.yml
the database used by yig storage backend should be created in docker-compose.yml.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hopkings2008:
This PR is to remove the old object from backend if the object is put with the same object key in the same bucket more than once.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #765 

**Release note**:

```release-note
```


s3/pkg/service/object.go
```diff
@@ -135,6 +135,12 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
 	obj.TenantId = tenantId
 	log.Infof("metadata of object is:%+v\n", obj)
 	log.Infof("*********bucket:%s,key:%s,size:%d\n", obj.BucketName, obj.ObjectKey, obj.Size)
+	oldObj, err := s.MetaStorage.GetObject(ctx, obj.BucketName, obj.ObjectKey, true)
```
sfzeng: Hi @hopkings2008 where the old object is deleted?

hopkings2008: @sfzeng when put the object with the same key used before, the underlining backend will remove the old one. the upper layer should not care about this. In s3 specification, when the put the same key, the old one will be overwritten. So, if the backend is yig, yig will responsible for removing the old one.

hopkings2008:
currently, this pr is depends on #744, so after that pr is merged, i can resolve the conflicts.

kumarashit:
#744 is merged, good to go for this?

s3/pkg/datastore/yig/storage/object.go
```diff
@@ -128,6 +128,22 @@ func (yig *YigStorage) Put(ctx context.Context, stream io.Reader, obj *pb.Object
...
 
+	// check and remove the old object if exists.
+	if obj.StorageMeta != "" && obj.ObjectId != "" {
+		storageMeta, err := ParseObjectMeta(obj.StorageMeta)
+		if err != nil {
+			log.Errorf("Put(%s, %s, %s) failed, failed to parse storage meta(%s), err: %v", obj.BucketName,
+				obj.ObjectKey, obj.ObjectId, obj.StorageMeta, err)
+			return result, err
+		}
+		err = yig.putObjToGc(storageMeta.Cluster, storageMeta.Pool, obj.ObjectId)
```
sfzeng: Hi @hopkings2008 , if versioning enabled, does different version of the same object have different objectId or not?

hopkings2008: Yes, if version is enabled, there are different oids for different versions. but the storage backend cannot see the version, it only handles oids. version is handled in upper layer.

sfzeng: OK, thanks.

hopkings2008: just transfer both the storage meta(transfer it even if it may be "", the backend will handle it) and object id to the backend.

hopkings2008:
> #744 is merged, good to go for this?

yes, it is fine to merge this pr.

PR #758, opened 2019-11-28
SSE changes for RC1

### LINKED ISSUE #634: PUT bucket SSE

rhsakarpos:
**Support SSE in OpenSDS. This PR adds support for the server generated random key encryption. The SSE-C and SSE-KMS are not targeted for DAITO**:

**Fixes the SSE related issue for DAITO**: fixes #634 #638 

**Release note**:

```The way this works is to encrypt the object byte stream in OSDS and upload to cloud bucket. Similarly, decrypt the byte stream on GET object (this is not done yet, waiting on GET to be implemented). The encryption settings on the cloud bucket are NOT altered. We use AES-256 with a random key, to encrypt the data before a PUT object happens

Further, in the current code, I am yet to get enough details on the following topics, will integrate my code with these once I have the information and complete the production testing with docker

1. AK-SK
2. Policies
3. Tenant ID

Tested the following on my local environment (without docker), using postman
1. PUT SSE option on bucket
2. GET SSE option on listing buckets
3. Encrypt data byte[ ] on PUT object
```


api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,172 @@
...
+var ClassAndTier1 map[string]int32
+var mutext1 sync.Mutex
+
+func (s *APIService) BucketSSEPut(request *restful.Request, response *restful.Response) {
+	bucketName := request.PathParameter("bucketName")
+	log.Infof("received request for PUT bucket SSE: %s", bucketName)
+
+	ctx := common.InitCtxWithAuthInfo(request)
+	bucket, err := s.s3Client.GetBucket(ctx, &s3.Bucket{Name: bucketName})
+	if err != nil {
```
sfzeng: Should we check both err and bucket.ErrorCode?

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		if sseConf.SSE.Enabled=="true"{
+			s3SSE.SseType = "SSE"
+		}
+
+		bucket.BucketMeta.ServerSideEncryption = s3SSE
+
+		s.s3Client.UpdateBucket(ctx, bucket.BucketMeta)
+
+		createLifecycleConf := model.LifecycleConfiguration{}
```
sfzeng: Lifecycle related codes need to be removed.

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,172 @@
...
+			XXX_unrecognized:     nil,
+			XXX_sizecache:        0,
+		}
+		if sseConf.SSE.Enabled=="true"{
+			s3SSE.SseType = "SSE"
+		}
+
+		bucket.BucketMeta.ServerSideEncryption = s3SSE
+
+		s.s3Client.UpdateBucket(ctx, bucket.BucketMeta)
```
sfzeng: What if UpdateBucket failed?

api/pkg/s3/listbuckets.go
```diff
@@ -45,7 +45,13 @@ func parseListBuckets(list *s3.ListBucketsResponse) []byte {
 				versionOpts.Status = utils.VersioningEnabled
 			}
 		}
-		bucket := model.Bucket{Name: value.Name, CreateTime: ctime, LocationConstraint: value.DefaultLocation, VersionOpts: versionOpts}
```
sfzeng: Is it better to provide a seperate API to get bucket SSE configuration, than return it in the List Bucket API?

api/pkg/s3/objectput.go
```diff
@@ -146,7 +149,13 @@ func (s *APIService) ObjectPut(request *restful.Request, response *restful.Respo
 			log.Debugln("finished read")
 			eof = true
 		}
-		err = stream.Send(&s3.PutObjectRequest{Data: buf[:n]})
+		// encrypt if needed
```
sfzeng: Is it better to do encryption in s3 service than in api service?

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -184,6 +192,14 @@ func (t *TidbClient) GetBuckets(ctx context.Context) (buckets []*Bucket, err err
 			tmp.Versioning.Status = versionOpts.Status
 		}
 
+		// get SSE info for this bucket
```
sfzeng: Is it necessary to return sse configuration when list buckets?

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -619,3 +635,61 @@ func (t *TidbClient) GetBucketVersioning(ctx context.Context, bucketName string)
...
+	}
+
+	return nil
+}
+
+func (t *TidbClient) GetBucketSSE(ctx context.Context, bucketName string) (sseOptsPtr *pb.ServerSideEncryption, err error) {
+	log.Info("list bucket SSE info from tidb ...")
+
+	var rows *sql.Rows
+	sqltext := "select sse from bucket_sseopts where bucketname=?;"
```
sfzeng: Does bucket_sseopts exist?

api/pkg/utils/encryptionutils.go
```diff
@@ -0,0 +1,78 @@
...
+	// https://en.wikipedia.org/wiki/Galois/Counter_Mode
+	aesgcm, gcmErr := cipher.NewGCM(cipherBlock)
+	if gcmErr != nil {
+		log.Errorf("Encryption error, GCM not generated")
+		return gcmErr, nil
+	}
+	nonce := make([]byte, aesgcm.NonceSize())
+	_, nonceErr := io.ReadFull(rand.Reader, nonce)
+	if nonceErr != nil {
+		log.Errorf("Encryption error, GCM nonce not createdf")
```
kumarashit:
```suggestion
		log.Errorf("Encryption error, GCM nonce not created")
```

api/pkg/utils/encryptionutils.go
```diff
@@ -0,0 +1,78 @@
...
+	// use the cipher block to encrypt
+	// gcm or Galois/Counter Mode, is a mode of operation
+	// for symmetric key cryptographic block ciphers
+	// https://en.wikipedia.org/wiki/Galois/Counter_Mode
+	aesgcm, gcmErr := cipher.NewGCM(cipherBlock)
+	if gcmErr != nil {
+		log.Errorf("Encryption error, GCM not generated")
+		return gcmErr, nil
+	}
+	nonce := make([]byte, aesgcm.NonceSize())
```
kumarashit: Should we comment that the NonceSize is default 12 bytes

kumarashit:
Liked the code

api/pkg/s3/router.go
```diff
@@ -46,4 +46,7 @@ func RegisterRouter(ws *restful.WebService) {
 	ws.Route(ws.DELETE("/{bucketName}/?lifecycle").To(handler.RouteBucketDelete)).Doc("Delete lifecycle configuration from the bucket")
 
 	ws.Route(ws.PUT("/{bucketName}/?versioning").To(handler.RouteBucketPut)).Doc("Create Versioning configuration for the bucket")
+
+	// router for SSE
+	ws.Route(ws.PUT("/{bucketName}/?DefaultEncryption").To(handler.RouteBucketPut)).Doc("Set default encryption on bucket")
```
hopkings2008: the url path should comply with the specification 'https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketEncryption.html'

api/pkg/utils/encryptionutils.go
```diff
@@ -0,0 +1,78 @@
...
+	// gcm or Galois/Counter Mode, is a mode of operation
+	// for symmetric key cryptographic block ciphers
+	// https://en.wikipedia.org/wiki/Galois/Counter_Mode
+	aesgcm, gcmErr := cipher.NewGCM(cipherBlock)
+	if gcmErr != nil {
+		log.Errorf("Decryption error, GCM not generated")
+		return gcmErr, nil
+	}
+	nonceSize := aesgcm.NonceSize()
+	nonce, ciphertext := data[:nonceSize], data[nonceSize:]
```
hopkings2008: do we need to check whether the length of data is more than the nonce size here?

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -112,6 +112,14 @@ func (t *TidbClient) GetBucket(ctx context.Context, bucketName string) (bucket *
 		tmp.Versioning.Status = versionOpts.Status
 	}
 
+	// get SSE info for this bucket
+	sseOpts, sseErr := t.GetBucketSSE(ctx, tmp.Name)
```
hopkings2008: it seems that a separate bucket sse table is used for bucket sse info?

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -619,3 +635,61 @@ func (t *TidbClient) GetBucketVersioning(ctx context.Context, bucketName string)
...
+		tmp := &pb.ServerSideEncryption{}
+		err = rows.Scan(
+			&tmp.SseType)
+		if err != nil {
+			err = handleDBError(err)
+			return
+		}
+		return tmp, nil
+	}
+	return
```
hopkings2008: before return, we should check rows.Error()

api/pkg/s3/listbuckets.go
rhsakarpos: hmm...this is used to list the buckets, SSE is one attribute of a bucket. So if we have 100 buckets, one call will return 100 bucket info, including SSE, instead of 200 calls, for example, to list the SSE column on dashboard. If you feel a separate API is ALSO needed, do revert here, i will add it

api/pkg/s3/objectput.go
rhsakarpos: better ? in what way ?

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: hmm...this is used to list the buckets, SSE is one attribute of a bucket. So if we have 100 buckets, one call will return 100 bucket info, including SSE, instead of 200 calls, for example, to list the SSE column on dashboard, for each bucket

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: yes, in the schema tidb.sql

api/pkg/s3/bucketsseput.go
```diff
@@ -0,0 +1,172 @@
...
+
+	ctx := common.InitCtxWithAuthInfo(request)
+	bucket, err := s.s3Client.GetBucket(ctx, &s3.Bucket{Name: bucketName})
+	if err != nil {
+		log.Errorf("get bucket failed, err=%v\n", err)
+		response.WriteError(http.StatusInternalServerError, fmt.Errorf("bucket does not exist"))
+	}
+
+	body := ReadBody(request)
+	log.Infof("MD5 sum for body is %x", md5.Sum(body))
```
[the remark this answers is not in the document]
rhsakarpos: done :-)

api/pkg/utils/encryptionutils.go
rhsakarpos: fixed

api/pkg/utils/encryptionutils.go
rhsakarpos: added

api/pkg/s3/router.go
rhsakarpos: hmm...i have a query here
```
why do we want to make all our APIs and header parameters etc. the same as what AWS currently supports ? specifically,
why only AWS ? why not HWS or GCP or Azure ?
i agree we need to functionally match-up, yes. but that can be done with different APIs and parameters
One more thing, say we do match AWS today, what is our mechanism to always keep it matching, to always be in sync with AWS ? What if AWS changes their API 2 weeks later, we keep playing catch-up ?
```

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: correct

api/pkg/utils/encryptionutils.go
rhsakarpos: not needed, the nonce is pre-pended to the encrypted data, so the data[] will be nonce-size + actual-data-size

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: added

rhsakarpos:
> Liked the code
Xie Xie ! i'll pass on the compliments to the original author.

api/pkg/s3/bucketsseput.go
rhsakarpos: added

api/pkg/s3/bucketsseput.go
rhsakarpos: done

api/pkg/s3/bucketsseput.go
rhsakarpos: handled error

api/pkg/s3/listbuckets.go
sfzeng: Ehmm.......I'm just thinking... if we need to want to be compatible with aws s3, as it's the de facto standard, then we need to provide a separate API.

api/pkg/s3/objectput.go
sfzeng: IMHO, API service only does parameter checking and request delivery, and s3 service will do encryption when received the request from api service.

api/pkg/s3/router.go
sfzeng: IMHO, currently, aws s3 is the de facto standard, there are many client tools or SDK which works based on s3 standard, if we do not compatible with s3 standard, that means those tools cannot work for us. So currently, we'd better be compatible with it, for the future, it depends.

api/pkg/utils/encryptionutils.go
hopkings2008: ok, if user uploads  some malformed data whose size is smaller than nonce-size, can we block this beforehand and make the program not crash?

api/pkg/s3/listbuckets.go
rhsakarpos: highly debatable...not convinced of the rationale in doing this

api/pkg/s3/objectput.go
rhsakarpos: raising issue #772 to track this

api/pkg/s3/router.go
rhsakarpos: tracked in #751

api/pkg/utils/encryptionutils.go
rhsakarpos: sure, raised issue to track this, will test and fix #773

PR #763, opened 2019-11-28
fix bug for issue762

### LINKED ISSUE #762: Lifecycle action does not take effect as supposed.

>
> /kind bug

**What happened**:
Object is not deleted or transit according the lifecycle rules.

Object can be deleted or transit according the lifecycle rules.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sfzeng:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #762 

**Release note**:

```release-note
```


datamover/pkg/drivers/https/prepare.go
```diff
@@ -200,92 +195,17 @@ func getOsdsS3Objs(ctx context.Context, in *pb.RunJobRequest, marker string, lim
...
-	objs, err := cephs3mover.ListObjs(defaultSrcLoca, filt)
-	if err != nil {
-		return nil, err
-	}
-	for i := 0; i < len(objs); i++ {
-		obj := osdss3.Object{Size: objs[i].Size, ObjectKey: objs[i].Key, Location: ""}
-		srcObjs = append(srcObjs, &obj)
+func GetMultipartSize() int64 {
+	var size int64 = 16 * 1024 * 1024 // this is the default
+	userSetSize, err := strconv.ParseInt(os.Getenv("PARTSIZE"), 10, 64)
```
kumarashit: Was thinking, should the PARTSIZE be coming from ENV or can it be an API to set the PARTSIZE?

s3/pkg/datastore/common/utils.go
```diff
@@ -12,16 +12,3 @@ func GetMd5FromCtx(ctx context.Context) (md5val string) {
 
 	return
 }
-
-func TrimQuot(in string) string {
```
kumarashit: Why removing this? Not used or will not be used?

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -229,7 +229,7 @@ func (t *TidbClient) PutBucket(ctx context.Context, bucket *Bucket) error {
 	lc, _ := json.Marshal(bucket.LifecycleConfiguration)
 	bucket_policy, _ := json.Marshal(bucket.BucketPolicy)
 	sql := "update buckets set bucketname=?,acl=?,policy=?,cors=?,lc=?,tenantid=?,versioning=? where bucketname=?"
-	args := []interface{}{bucket.Name, acl, bucket_policy, cors, lc, bucket.TenantId, bucket.Versioning, bucket.Name}
+	args := []interface{}{bucket.Name, acl, bucket_policy, cors, lc, bucket.TenantId, bucket.Versioning.Status, bucket.Name}
```
kumarashit: We fixed this part earlier, right?

datamover/pkg/drivers/https/prepare.go
hopkings2008: In my opinion, the size of part should be got from configuration file or from db? if it is a configurable parameter, should it be put into the configuration file?

sfzeng: Right now, it's from ENV, see os.Getenv("PARTSIZE"), it's setted in docker-compose.yml.

s3/pkg/datastore/common/utils.go
sfzeng: Will not be used, cause etag got from AWS s3, Azure Blob, etc, start and end with double quotation marks, we will keep the same as them.

s3/pkg/meta/db/drivers/tidb/bucket.go
sfzeng: This is another one, the one we fixed 3 days ago are for creating bucket, this one is for updating bucket.

s3/pkg/datastore/common/utils.go
sfzeng: Sorry, it should not be removed, I correct it now, thanks.

datamover/pkg/drivers/https/prepare.go
kumarashit: Ok. My opinion was to make the PARTSIZE setting to be configurable from API so that any application using multi-cloud can get the leverage rather than to change YAML or ENV. But that will be an ehancement

sfzeng: OK, let me create an enhance issue to trace it.

sfzeng: Here is the issue: [#771 ](https://github.com/opensds/multi-cloud/issues/771)

PR #766, opened 2019-11-29
return last modify time in list parts interface of storage backend

hopkings2008:
Return last modify time of parts in list parts interface of storage backend, currently, only yig returns this element, the other backend will changes in later pr.

**Release note**:

```release-note
```


s3/pkg/datastore/yig/storage/multipart.go
```diff
@@ -256,9 +256,10 @@ func (yig *YigStorage) ListParts(ctx context.Context, multipartUpload *pb.ListPa
 			break
 		}
 		p := model.Part{
-			PartNumber: part.PartNum,
-			ETag:       part.Etag,
-			Size:       int64(part.Size),
+			PartNumber:     part.PartNum,
+			ETag:           part.Etag,
+			Size:           int64(part.Size),
+			LastModifyTime: part.UpdateTime.Unix(),
```
kumarashit: Doubt..When listing the parts, the last modify time should be based upon the existing value from DB and we should not set the LastModifyTime while listing..Isn't it?

hopkings2008: @kumarashit yes it is. The last modify time is created when the part is uploaded successfully. And we should not set it when listing it. And in this pr, the last modify time of the part is not changed in ListParts interface, it is got from DB.

PR #769, opened 2019-12-03
Support the API of list object part

### LINKED ISSUE #768: Support the API of ListPart and CopyPart
> /kind feature

ListPart API:  Lists the parts that have been uploaded for a specific multipart upload,  the requirement is from https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html
CopyPart API:  Uploads a part by copying data from an existing object as data source. the requirement is from https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html.  Precondition feature for CopyPart is supported in the  PR

the features are not supported in the CopyPart:
sse
version

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:
Support the API of **CopyObjectPart** and **ListObjectPart**

ListPart API: Lists the parts that have been uploaded for a specific multipart upload, the requirement is from https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html.
CopyPart API: Uploads a part by copying data from an existing object as data source. the requirement is from https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html. Precondition feature for CopyPart is supported in the PR

the features are not supported in the CopyPart:
sse
version

**Release note:**

 I only test the two APIS on YIG backend, and in next PR, we will add uploadpart table in tidb schema to support other backend.


s3/pkg/service/multipart.go
```diff
@@ -444,8 +444,202 @@ func (s *s3Service) AbortMultipartUpload(ctx context.Context, in *pb.AbortMultip
...
+		log.Errorln("get bucket failed with err:", err)
+		return err
+	}
+	isAdmin, tenantId, err := util.GetCredentialFromCtx(ctx)
+	if err != nil && isAdmin == false {
+		log.Error("get tenant id failed")
+		err = ErrInternalError
+		return err
+	}
+	if !isAdmin {
```
sfzeng: Do we need to check ACL of both target bucket and source bucket, and source object maybe?

s3/pkg/service/multipart.go
```diff
@@ -444,8 +444,202 @@ func (s *s3Service) AbortMultipartUpload(ctx context.Context, in *pb.AbortMultip
...
+		switch multipart.Metadata.Acl.CannedAcl {
+		case "public-read", "public-read-write":
+			break
+		case "authenticated-read":
+			if tenantId == "" {
+				err = ErrAccessDenied
+				return err
+			}
+		case "bucket-owner-read", "bucket-owner-full-controll":
```
sfzeng: Does CannedAcl  include bucket-owner-read and bucket-owner-full-controll? Isn't owner always have full control?

s3/pkg/service/multipart.go
sunfch: @sfzeng   I added common function to check get object acl at the pr of "valid acl rules", so the source object acl check will be added after it is merged.  I add a new issue to track.

s3/pkg/service/multipart.go
sunfch: @sfzeng  yes, these two rules are member of CannedAcl.

s3/pkg/service/multipart.go
sfzeng: OK, thanks.

s3/pkg/service/multipart.go
sfzeng: Ok, thanks.

sfzeng:
Hi @sunfch it seems not all kinds of backend implemented the ListPart API, so could you please add an isse to trace it and add fix it later?

sunfch:
> Hi @sunfch it seems not all kinds of backend implemented the ListPart API, so could you please add an isse to trace it and add fix it later?

I have add a new issue #781  to track

sfzeng:
@sunfch OK, thanks, and .... a little suggestion, next time, please describe the details of the question in issue , instead of coping comment directly :-)

sfzeng:
@sunfch Thanks.

skdwriting:
@sunfch @sfzeng , let us clearly list the supported backends.

api/pkg/s3/api-response.go
```diff
@@ -194,4 +201,38 @@ func parseListUploadsQuery(query url.Values) (request ListUploadsRequest, err er
...
+
+// Parse object url queries
+func parseListObjectPartsQuery(query url.Values) (request ListPartsRequest, err error) {
+	request.EncodingType = query.Get("encoding-type")
+	request.UploadId = query.Get("uploadId")
+	if request.UploadId == "" {
+		err = ErrNoSuchUpload
+		return
+	}
+	if query.Get("max-parts") == "" {
```
skdwriting: if its null, we default to MaxPartsList ? -- is this correct? Because other invalid checks are handled below.

api/pkg/s3/objectpartcopy.go
```diff
@@ -0,0 +1,170 @@
...
+
+package s3
+
+import (
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/emicklei/go-restful"
+	"github.com/journeymidnight/yig/helper"
```
skdwriting: as this helper and such functions become core part within multicloud as native integration, should we move the helper code to multicloud github as package and then import?

skdwriting: there is already helper in https://github.com/opensds/multi-cloud/s3/pkg/helper

sfzeng: @skdwriting You are right, this should be s3/pk/helper.

sfzeng: @sunfch Could you please take a look of this?

api/pkg/s3/api-response.go
sfzeng: I thinks this max-parts is used to sets the maximum number of parts to return, if the user did not set it, then we use a default value, for example, we return 1000 parts as default.

api/pkg/s3/objectpartcopy.go
sunfch: @skdwriting @sfzeng , sorry for my mistake.  I fix it right now, thanks

sunfch:
> @sunfch @sfzeng , let us clearly list the supported backends.

Thanks for your reviewing
The YIG-Ceph backend is only supported in the PR.  I will submit a new PR to support aws backend, and a new issue #781 have been added to track,  and I also add it to the RN of the PR

api/pkg/s3/api-response.go
sunfch: @skdwriting  thanks for your reviewing,  if max-parts is not specified by user, we will use MaxPartsList as default.

sunfch:
I will separate the two APIS of the PR, and submit a new PR that support only one API,  thanks

PR #770, opened 2019-12-03
fix the problem of extra pressure to ceph clusters made by concurrent running yig storage backends.

### LINKED ISSUE #759: GC will be performed at the same time of yig storage backend and it will introduce more pressure to db and ceph cluster.
This is a bug report and the bug is about gc will be performed at the same time of yig storage backend and it will introduce more pressure to db and ceph cluster.
>
> /kind bug
> /kind feature

**What happened**:
if multiple yig storage backend instances run at the same time, gc will be executed at the same time and this will introduce the extra pressure to the db and ceph cluster due to one of them will really perform the work and the other doesn't work actually because the data has already been removed by one of them.
if multiple yig storage backend starts at the same time and they all access the same db and ceph cluster, gc of them will remove the different set of garbage objects.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hopkings2008:
…the same time which will introduce un-needed pressure for ceph clusters.

This PR fixes the problem of introducing extra pressure to ceph clusters made by concurrent running yig storage backends.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #759 

**Release note**:

```release-note
```


s3/pkg/datastore/yig/storage/gc.go
```diff
@@ -70,6 +79,22 @@ func (gm *GcMgr) Start() {
 				}
 				log.Errorf("failed to remove object: %s, err: %s", result.ObjectId, result.Err)
 			}
+			gcEnd := time.Now()
```
sfzeng: Hi @hopkings2008 Could you please add comment to explain the logic here? For example, what count and  backoffCound stand for, why need intervalTime and maxIntervalTime?

hopkings2008: @sfzeng thanks and i have added more comments for this logic, pls check it again.

s3/pkg/datastore/yig/storage/gc.go
```diff
@@ -16,6 +16,7 @@ package storage
 import (
```
rhsakarpos: 'gc' means ? Garbage Collection ?

s3/pkg/datastore/yig/storage/gc.go
```diff
@@ -27,32 +28,47 @@ import (
...
+	loopTime int64
+	wg       sync.WaitGroup
 }
 
 func (gm *GcMgr) Start() {
 	// query the current available cpus
 	threadNum := runtime.GOMAXPROCS(0)
 	gm.wg.Add(1)
 	go func() {
+		// loopCount is the number of loops for performing gc.
```
rhsakarpos: this looks tricky, please share the requirement and design doc please, thanks

s3/pkg/datastore/yig/storage/gc.go
hopkings2008: yes, it is

hopkings2008:
> Question...why would OSDS run the YIG backend ? that is running someplace outside of OSDS, right ? like any other cloud backend ? or am i missing something ?
The level of YIG backend in OSDS is the same as aws backend. and yes, it is running outside of OSDS, it is only controlled by yig backend like any other backend.
Please don't confuse this with the gc in OSDS, they are two different things.

s3/pkg/datastore/yig/storage/gc.go
hopkings2008: this was already discussed in slack before, also, shufang had sent the design pictures in that conversation. pls check the conversation in the slack, thanks.

PR #774, opened 2019-12-03
Support migration based on CopyObject API provided by OSDS S3 service

sfzeng:
This PR refactored migration function based on the CopyObject API provided OSDS S3 service, that means datamover do not need to migrate data by itself, then we do not need to maintain two copies of code for the same function. 

As those codes before used for migration data in datamover are removed, the progress update also need to do some change, so @Click2Cloud-Gamma please review the part about progress update.

**Release note**:

```release-note
With this PR, objects less than PARTSIZE(16M as default) can be migrated, while objects bigger than this size can not be migrated, for those bigger objects, the next PR will provide the migration ability.
```

### DIFF
--- a/datamover/pkg/drivers/https/migration.go
+++ b/datamover/pkg/drivers/https/migration.go
@@ -290,7 +290,7 @@ func runjob(in *pb.RunJobRequest) error {
 				log.Infof("ObjectMigrated:%d,TotalCapacity:%d Progress:%d\n", j.PassedCount, j.TotalCapacity, j.Progress)
 				db.DbAdapter.UpdateJob(&j)
 			}
-		case <-time.After(time.Duration(JOB_RUN_TIME_MAX) * time.Second):
+		case <-time.After(time.Duration(dur) * time.Second):
 			{
 				tmout = true
 				log.Warnln("Timout.")
--- a/datamover/pkg/drivers/https/migration.go
+++ b/datamover/pkg/drivers/https/migration.go
@@ -42,8 +42,6 @@ var JOB_RUN_TIME_MAX = 86400           //seconds, equals 1 day
 var s3client osdss3.S3Service
 var bkendclient backend.BackendService
 
-const WT_DOWLOAD = 48
-const WT_UPLOAD = 48
 const WT_MOVE = 96
 const WT_DELETE = 4
 const JobType = "migration"
@@ -266,7 +264,6 @@ func runjob(in *pb.RunJobRequest) error {
 		}
 
 		//Do migration for each object.
-		//go doMove(ctx, objs, capa, th, srcLoca, destLoca, in.RemainSource, &j)
 		go doMigrate(ctx, objs, capa, th, in, &j)
 
 		if num < int(limit) {
--- a/datamover/pkg/drivers/https/migration.go
+++ b/datamover/pkg/drivers/https/migration.go
@@ -18,9 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"log"
 	"math"
-	"os"
 	"strconv"
 	"time"
 
@@ -30,17 +28,12 @@ import (
 	"github.com/opensds/multi-cloud/api/pkg/common"
 	"github.com/opensds/multi-cloud/backend/proto"
 	flowtype "github.com/opensds/multi-cloud/dataflow/pkg/model"
-	"github.com/opensds/multi-cloud/datamover/pkg/amazon/s3"
-	"github.com/opensds/multi-cloud/datamover/pkg/azure/blob"
-	"github.com/opensds/multi-cloud/datamover/pkg/ceph/s3"
 	"github.com/opensds/multi-cloud/datamover/pkg/db"
-	Gcps3mover "github.com/opensds/multi-cloud/datamover/pkg/gcp/s3"
-	obsmover "github.com/opensds/multi-cloud/datamover/pkg/huawei/obs"
-	ibmcosmover "github.com/opensds/multi-cloud/datamover/pkg/ibm/cos"
 	. "github.com/opensds/multi-cloud/datamover/pkg/utils"
 	pb "github.com/opensds/multi-cloud/datamover/proto"
 	s3utils "github.com/opensds/multi-cloud/s3/pkg/utils"
 	osdss3 "github.com/opensds/multi-cloud/s3/proto"
+	log "github.com/sirupsen/logrus"
 )
 
 var simuRoutines = 10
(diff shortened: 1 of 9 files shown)


datamover/pkg/drivers/https/migration.go
```diff
@@ -49,19 +42,19 @@ var JOB_RUN_TIME_MAX = 86400           //seconds, equals 1 day
 var s3client osdss3.S3Service
 var bkendclient backend.BackendService
 
-var logger = log.New(os.Stdout, "", log.LstdFlags)
-
 const WT_DOWLOAD = 48
 const WT_UPLOAD = 48
```
Click2Cloud-Gamma: @sfzeng Since we don't have sub-process like  download and upload here in migration.go, so this 2 constant `WT_DOWNLOAD` and `WT_UPLOAD` should be removed.

datamover/pkg/drivers/https/migration.go
```diff
@@ -615,7 +266,9 @@ func runjob(in *pb.RunJobRequest) error {
 		}
 
 		//Do migration for each object.
-		go doMove(ctx, objs, capa, th, srcLoca, destLoca, in.RemainSource, &j)
+		//go doMove(ctx, objs, capa, th, srcLoca, destLoca, in.RemainSource, &j)
```
Click2Cloud-Gamma: please remove this commented old code. Thanks

datamover/pkg/drivers/https/migration.go
sfzeng: Done, thanks !

datamover/pkg/drivers/https/migration.go
sfzeng: Done, thanks !

datamover/pkg/drivers/https/prepare.go
```diff
@@ -9,143 +9,44 @@ import (
...
 )
 
-func getCtxTimeout() time.Duration {
-	// 1 day as default
-	tmout := 86400 * time.Second
+func GetCtxTimeout(key string, min, max, def int64) time.Duration {
+	// default value
+	tmout := time.Duration(def) * time.Second
 
-	tmoutCfg, err := strconv.ParseInt(os.Getenv("JOB_MAX_RUN_TIME"), 10, 64)
```
rhsakarpos: this is good, can be made generic like this ? 
`
func GetEnvVar(key string, min, max, def int64, varName string) {}
`
and use like this
`
getEnvVar(key, 10, 200, "JOB_MAX_RUN_TIME")
`

datamover/pkg/drivers/lifecycle/incloud_transition.go
```diff
@@ -51,7 +51,10 @@ func doInCloudTransition(acReq *datamover.LifecycleActionRequest) error {
 		return errors.New(DMERR_TransitionInprogress)
 	}
 
-	ctx, _ := context.WithTimeout(context.Background(), CLOUD_OPR_TIMEOUT*time.Second)
+	// min is 1 minute, max is 30 days, default is 1 hour, user defined value should not less than min and not more than
+	// max, otherwise default will be used
+	tmout := migration.GetCtxTimeout("OBJECT_MOVE_TIME", SECONDS_ONE_MINUTE, SECONDS_30_DAYS, SECONDS_ONE_HOUR)
+	ctx, _ := context.WithTimeout(context.Background(), tmout)
```
rhsakarpos: question here, say the timeout expires, all files are NOT yet moved...is there a persisted log someplace or any events on UI, to show this to the user, to retry the incomplete items ?

rhsakarpos: my concern is this...recovering from a timed out timeout is usually messy. Further, arriving at an optimal timeout is also not easy. Should we do something like an auto-extend-timeout option ? How it will work is 

1. say you have a timeout of 30 seconds
2. 5 files need to be moved from A to B
3. at run time, we know 3 files have been moved and 20 seconds have elapsed (we will run out of time)
4. now, we auto-extend the timeout by 15 more seconds (to allow task to finish) if the user had selected this auto-extend option

sfzeng: Hi @rhsakarpos There are two scenarios related, lifecycle management and migration. For lifecycle management, if it failed, we just print an warn log and it will try again in the next round, lifecycle scheduler runs periodically. For migration, the job will return and set the status to be failed.

sfzeng: Thanks for your suggestion, for auto-extend-timeout, I think we can do it as an enhancement in the future. Currently, for migration, if it failed, user can run it again. For lifecycle management, it will try again in the next schedule round automatically.

datamover/pkg/drivers/https/prepare.go
sfzeng: Hi @rhsakarpos isn't key the same as varName?

datamover/pkg/drivers/https/migration.go
```diff
@@ -49,19 +42,17 @@ var JOB_RUN_TIME_MAX = 86400           //seconds, equals 1 day
 var s3client osdss3.S3Service
 var bkendclient backend.BackendService
 
-var logger = log.New(os.Stdout, "", log.LstdFlags)
-
-const WT_DOWLOAD = 48
-const WT_UPLOAD = 48
+const WT_MOVE = 96
```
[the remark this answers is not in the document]
sfzeng: Hi @hopkings2008  the former two are already removed according to Click2Cloud-Gamma's suggestion. Thanks.

datamover/pkg/drivers/https/prepare.go
rhsakarpos: oh...yes it is. so inside this function, there is no specific key like "JOB_MAX_RUN_TIME"

rhsakarpos: updated my initial comment to reflect this

datamover/pkg/drivers/lifecycle/incloud_transition.go
rhsakarpos: sounds good. However, will migration also list which objects were migrated and which were not ?

rhsakarpos: ok

sfzeng: No, not list now, it's only print in log, maybe we can add this part in the future in some way, cause objects the number of objects maybe huge.

datamover/pkg/drivers/https/prepare.go
sfzeng: Yes, it is generic as what you said.

datamover/pkg/drivers/lifecycle/incloud_transition.go
rhsakarpos: sure, log is also good, something for the user to see and act on

PR #776, opened 2019-12-04
Valid acl in all api and improve PutObject API

### LINKED ISSUE #775: Valid Canned ACL

> /kind feature

valid acl rule in s3 API and improve PutObject API

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:

The PR has some features and improvement as follows:
1. valid Canned Acl rules in S3 API.  ONLY "private" and "public-read" and "public-read-write" Canned Acl rules are included in the PR, and Other Canned Acls only are not supported. We can improve the Canned Acls later. The requirement is from https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
  The "private" and "public-read" and "public-read-write" Canned Acl can all applied to bucket and object.
  Their Permissions are as follows:
  private: Owner gets FULL_CONTROL. No one else has access rights (default).
  public-read: Owner gets FULL_CONTROL. The AllUsers group gets READ access
  public-read-write: Owner gets FULL_CONTROL. The AllUsers group gets READ and WRITE access. Granting this on a bucket is generally not recommended
  

2. improve PutObject API. delete unused code and implement incomplete function

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #775 

**Release note**:

```release-note
```
__


sfzeng:
Hi @sunfch Could you please add a comment to explain what kinds of ACL we support now, and with each kind of ACL, what operations can be done by those user who do not own the resource(bucket, object, etc.)?

s3/pkg/service/bucket.go
```diff
@@ -125,13 +125,22 @@ func (s *s3Service) GetBucket(ctx context.Context, in *pb.Bucket, out *pb.GetBuc
...
 	if err != nil {
-		if err == ErrNoSuchKey {
-			err = ErrNoSuchBucket
+		log.Errorf("get credential faied, err:%v\n", err)
+		return err
+	}
+	if !isAdmin {
+		if tenantId != bucket.TenantId {
+			switch bucket.Acl.CannedAcl {
+			case "public-read", "public-read-write", "authenticated-read":
```
sfzeng: What about private ACL?

s3/pkg/service/bucket.go
```diff
@@ -165,11 +174,9 @@ func (s *s3Service) DeleteBucket(ctx context.Context, in *pb.Bucket, out *pb.Bas
 		log.Errorf("get bucket failed, err:%+v\n", err)
 		return nil
 	}
-	_, _, err = CheckRights(ctx, bucket.TenantId)
+	_, _, _, err = CheckRights(ctx, bucket.TenantId)
```
sfzeng: Does the users other than the owner have the rights to delete a bucket?

s3/pkg/service/bucket.go
```diff
@@ -237,11 +242,9 @@ func (s *s3Service) GetBucketLifecycle(ctx context.Context, in *pb.BaseRequest,
 		return nil
 	}
 
-	_, _, err = CheckRights(ctx, bucket.TenantId)
+	_, _, _, err = CheckRights(ctx, bucket.TenantId)
 	if err != nil {
```
sfzeng: Here need to return ErrBucketAccessForbidden,  otherwise the user may create a new one with the same name, but it will failed because duplicate key.

s3/pkg/service/bucket.go
```diff
@@ -267,11 +270,9 @@ func (s *s3Service) DeleteBucketLifecycle(ctx context.Context, in *pb.BaseReques
 		log.Errorf("get bucket err: %v\n", err)
 		return nil
 	}
-	_, _, err = CheckRights(ctx, bucket.TenantId)
+	_, _, _, err = CheckRights(ctx, bucket.TenantId)
 	if err != nil {
-		if err == ErrNoSuchKey {
```
sfzeng: Same here.

s3/pkg/service/bucket.go
```diff
@@ -330,18 +331,10 @@ func (s *s3Service) PutBucketACL(ctx context.Context, in *pb.PutBucketACLRequest
...
-		return err
-	}
-
-	if isAdmin == false {
-		if bucket.TenantId != tenantId {
-			err = ErrBucketAccessForbidden
-			return err
-		}
+	_, _, _, err = CheckRights(ctx, bucket.TenantId)
+	if err != nil {
```
sfzeng: Same here

s3/pkg/service/multipart.go
```diff
@@ -46,7 +46,7 @@ func (s *s3Service) ListBucketUploadRecords(ctx context.Context, in *pb.ListBuck
 		return err
```
hopkings2008: pls set the err to the out so that upper framework can detect it.

s3/pkg/service/bucket.go
```diff
@@ -207,11 +214,9 @@ func (s *s3Service) PutBucketLifecycle(ctx context.Context, in *pb.PutBucketLife
 		return nil
 	}
 
-	_, _, err = CheckRights(ctx, bucket.TenantId)
+	_, _, _, err = CheckRights(ctx, bucket.TenantId)
 	if err != nil {
-		if err == ErrNoSuchKey {
```
sfzeng: Is it better to return ErrBucketAccessForbidden here,  otherwise the user may create a new one with the same name, but it will failed because duplicate key.

s3/pkg/service/object.go
```diff
@@ -296,8 +262,9 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
 		log.Errorln("failed to get object info from meta storage. err:", err)
 		return err
 	}
-	_, _, err = CheckRights(ctx, object.TenantId)
```
sfzeng: Does the ACL matters here?

s3/pkg/service/object.go
```diff
@@ -93,165 +92,131 @@ func (dr *StreamReader) Read(p []byte) (n int, err error) {
...
-	obj.ContentType = md["Content-Type"]
+	obj.ContentType = req.ContentType
 	obj.DeleteMarker = false
-	obj.CustomAttributes = md /* TODO: only reserve http header attr*/
+	obj.CustomAttributes = req.Attrs
 	obj.Type = meta.ObjectTypeNormal
 	obj.Tier = utils.Tier1 // Currently only support tier1
 	obj.StorageMeta = res.Meta
+	obj.Size = req.Size
+	obj.Location = req.Location
```
sfzeng: The req.Location can be null, if it is null, then so obj.Location should be the default location of the bucket.

s3/pkg/service/object.go
```diff
@@ -93,165 +92,131 @@ func (dr *StreamReader) Read(p []byte) (n int, err error) {
...
-	res, err := sd.Put(ctx, limitedDataReader, obj)
+	res, err := sd.Put(ctx, limitedDataReader, &pb.Object{BucketName:req.BucketName, ObjectKey:req.ObjectKey})
 	if err != nil {
 		log.Errorln("failed to put data. err:", err)
 		return err
 	}
-	oid := res.ObjectId
-	bytesWritten := res.Written
+	var delObj *pb.DeleteObjectInput
+	defer s.removeObjectFromBackend(ctx, sd, delObj)
```
sfzeng: Is it better to call the function directly instead of using defer?

s3/pkg/service/object.go
```diff
@@ -429,6 +397,30 @@ func (s *s3Service) CopyObject(ctx context.Context, in *pb.CopyObjectRequest, ou
...
+			break
+		default:
+			if targetBucket.TenantId != tenantId {
+				err = ErrBucketAccessForbidden
+				return err
+			}
+		}
+	}
+
 	srcBucket, err := s.MetaStorage.GetBucket(ctx, srcBucketName, true)
```
sfzeng: Is it necessary to check the access rights of source bucket and source object?

sunfch:
> Hi @sunfch Could you please add a comment to explain what kinds of ACL we support now, and with each kind of ACL, what operations can be done by those user who do not own the resource(bucket, object, etc.)?

@sfzeng   I have added it in RN

s3/pkg/service/bucket.go
sunfch: @sfzeng  I tested it with aws, and found that only owner can delete a bucket.

s3/pkg/service/multipart.go
sunfch: @hopkings2008 thanks,  will do it

s3/pkg/service/object.go
sunfch: @sfzeng , I checked the code again, and found the CheckRights here is not necessary, and I will delete it

s3/pkg/service/object.go
sunfch: @sfzeng  thanks, will fix it

s3/pkg/service/object.go
sunfch: @sfzeng , if there are more than one error where we need delete written object, it is better to use defer?

s3/pkg/service/object.go
sfzeng: Hi @sunfch What do you mean by 'not necessary'? Isn't it necessary to check ACL when get object?

s3/pkg/service/object.go
sunfch: @sfzeng  thanks, you are right, I will add the check in next commit

s3/pkg/service/bucket.go
sfzeng: OK, thanks.

hopkings2008:
LGTM

s3/pkg/service/object.go
sunfch: @sfzeng  I found that CheckRights is called twice here, so I delete one

skdwriting:
@sfzeng , please complete the review and approve. Thanks

skdwriting:
@sunfch this is tested with all currently supported backends in opensds?

PR #780, opened 2019-12-05
support to clean uncompleted multiparts in lifecycle

sfzeng:
This PR is to provide the ability to clean uncompleted multipart in lifecycle management. That is the user can set lifecyle rules to clean uncompleted parts some days after it's creation.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #780 

**Release note**:

```release-note
As Azure blob does not support to abort any multipart, so this rule does not work for those objects which stored in Azure Blob. Instead, Azure will clean those uncompleted multiparts 7 days after it's creation automatically. 
```


hopkings2008:
LGTM

PR #786, opened 2019-12-06
PUT object with versioning

rhsakarpos:
A quick description of how we plan on supporting object versions on PUT object in OSDS, please check if this will hold up. In OSDS, we will use the versionId field in the object table
on cloud backend, we will update the object key to include the OSDS version Id (this can be used to GET specific version etc.)

image

image

**Release note**:

```release-note
```


rhsakarpos:
@sfzeng @kumarashit @nguptaopensds please review, thanks

s3/pkg/meta/db/drivers/tidb/gc.go
```diff
@@ -94,21 +91,20 @@ func (t *TidbClient) ListGcObjs(ctx context.Context, offset, limit int) (objs []
...
 			&obj.BucketName,
 			&obj.ObjectKey,
-			&iversion,
+			&obj.VersionId,
 			&obj.Location,
 			&obj.ObjectId,
 			&obj.StorageMeta,
 		)
-		timestamp := math.MaxUint64 - iversion
+		/*timestamp := math.MaxUint64 - iversion
```
sfzeng: @sunfch @hopkings2008 Please check this part, is there any problem to change like this?

s3/pkg/service/object.go
```diff
@@ -164,6 +165,18 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
 		return err
 	}
+	// TODO Prakash(rhsakarpos@gmail.com) once Neelam checks in the GET versioning on bucket code, add it here
+	isBucketVersioned := true
+	if isBucketVersioned{
+		versionId, now := utils.GetVersionId()
+		obj.VersionId = versionId
+		obj.LastModified = now
+		// update the cloud backend object name here
+		obj.ObjectKey = obj.ObjectKey + "_" + strconv.FormatUint(obj.VersionId,10)
```
sfzeng: Does this will conclude a different object key in database? Actually we have obj.objectId which is used to store the key used in cloud backend.

s3/pkg/utils/utils.go
```diff
@@ -134,3 +136,13 @@ func GetBackend(ctx context.Context, backedClient backend.BackendService, backen
 	backend := backendRep.Backends[0]
 	return backend, nil
 }
+
+func GetVersionId() (versionId uint64, timeStamp int64) {
+	now := time.Now().UTC().Unix()
+	version := math.MaxUint64 - uint64(now)
```
sfzeng: If we use this kind of versionId, is it easy for conflict to happen?

sfzeng: IMHO, it's better to use nanoseconds instead of seconds, otherwise conflict may be easy to happen.

sfzeng: And even no need to use math.MaxUint64 - now, but use now it self.

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -118,7 +118,11 @@ func (t *TidbClient) GetBucket(ctx context.Context, bucketName string) (bucket *
 		return
 	}
 	tmp.ServerSideEncryption = &pb.ServerSideEncryption{}
-	tmp.ServerSideEncryption.SseType = sseOpts.SseType
+	if sseOpts!=nil{
```
kumarashit: Duplicate of PR#787?

s3/pkg/service/object.go
sfzeng: Hi @rhsakarpos please check this, cause different object key means a different object, I guess this is not supposed to be here, is it?

rhsakarpos: yes, different version = different object, sorry, did not get the problem...

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: no...this is a nil check added

s3/pkg/meta/types/object.go
```diff
@@ -186,7 +184,7 @@ func (o *Object) encryptSseKey() (err error) {
 	return nil
 }
 
-func (o *Object) GetVersionId() string {
+/*func (o *Object) GetVersionId() string {
```
kumarashit: Can we remove this commented code? Anyways, this will be coming in another PR, right?

s3/pkg/service/object.go
sfzeng: For example, user put object twice which use obj as it's key, and then user will have two versions of the same key 'obj', not obj_xxxxxx and obj_yyyyy, right?

rhsakarpos:
closing, since could not merge  > 10 days

PR #787, opened 2019-12-07
SSE move PUT changes from API to S3 layer

rhsakarpos:
Moves the code for encryption during PUT object, to the S3 layer, to enable re-use during migration and copy of object

**Release note**:

image

```

```


rhsakarpos:
@sfzeng @nguptaopensds @kumarashit please review, thanks

s3/pkg/service/object.go
```diff
@@ -155,6 +158,15 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
 		limitedDataReader = data
 	}
 
+	// encrypt if needed
+	if bucket.ServerSideEncryption.SseType == "SSE" {
+		byteArr, _ := ioutil.ReadAll(limitedDataReader)
+		key, _ := utils.GetRandom32BitKey()
+		_, encBuf := utils.EncryptWithAES256RandomKey(byteArr, key)
```
sfzeng: Hi @rhsakarpos Does this mean it will read all of the input and put it to a buffer, and then encrypt it and save to a new buffer? If it is like that, then that would need too much memory.

rhsakarpos: yes @sfzeng right, it will take double the initial memory. opening issue to track this, will implement your idea of a wrapped reader, which should do this ?
_The wrapped reader looks good, i think this will save memory ? as in, it will read and encrypt on the fly ? as against read-all-make-a-copy-then-encrypt-all ?_
Same goes for decrypt too, as mentioned in #788

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -640,11 +646,11 @@ func (t *TidbClient) GetBucketVersioning(ctx context.Context, bucketName string)
 	return
 }
 
-func (t *TidbClient) CreateBucketSSE(ctx context.Context, bucketName string, sseType string) error {
+func (t *TidbClient) CreateBucketSSE(ctx context.Context, bucketName string, sseType string, sseKey []byte) error {
 	log.Infof("create bucket[%s] SSE info[%s] into tidb ...\n", bucketName, sseType)
 
-	sql := "insert into bucket_sseopts(bucketname, sse) values(?,?);"
```
sfzeng: Hi @rhsakarpos Can we save SSE settings in buckets table instead of bucket_sseopts?

rhsakarpos: will do @sfzeng, i believe RC2 was yesterday or is today, in which case can this be done after the RC2 cut ?

sfzeng: OK, how about creating an issue to trace it?

rhsakarpos: here goes #792

s3/pkg/service/object.go
```diff
@@ -362,6 +375,20 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
 			break
 		}
 
+		if bucket.ServerSideEncryption.SseType == "SSE" {
+			// decrypt and write
+			fmt.Println("from cloud")
+			fmt.Println(buf[0:n])
```
sfzeng: Here, use log instead of fmt please, and is it a good idea to print all the buffer?

s3/pkg/service/object.go
```diff
@@ -362,6 +375,20 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
+			decErr, decBytes := utils.DecryptWithAES256(buf[0:n], bucket.ServerSideEncryption.EncryptionKey)
+			if decErr != nil {
+				log.Errorln("failed to decrypt data. err:", decErr)
+				return decErr
+			}
+			fmt.Println("successfully decrypted data")
```
sfzeng: Same here.

s3/pkg/service/object.go
rhsakarpos: apologies @sfzeng, was debugging some byte padding issue locally, forgot to remove these before commit, removed now

s3/pkg/service/object.go
rhsakarpos: changed to log

kumarashit:
Associated PR to track #800

PR #797, opened 2019-12-11
Support the API of list object parts

sunfch:

The PR support the API of list object parts for in-progress multipart uploads.  

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #769 

I add a objectpart table in tidb schema to save object part,  and delete the object part items in the table when completemultipart  is called.   If there is no the table, we cannot get total object size or etag from some kind of backends. For example, we cannot get object etag from azure backend.

**Release note**:

```
```


api/pkg/s3/api-response.go
```diff
@@ -194,4 +194,37 @@ func parseListUploadsQuery(query url.Values) (request ListUploadsRequest, err er
...
+
+// Parse object url queries
+func parseListObjectPartsQuery(query url.Values) (request ListPartsRequest, err error) {
+	request.EncodingType = query.Get("encoding-type")
+	request.UploadId = query.Get("uploadId")
+	if request.UploadId == "" {
+		err = ErrNoSuchUpload
+		return
+	}
+	if query.Get("max-parts") == "" {		request.MaxParts = MaxPartsList
```
sfzeng: Format the code please.

api/pkg/s3/api-response.go
```diff
@@ -194,4 +194,37 @@ func parseListUploadsQuery(query url.Values) (request ListUploadsRequest, err er
...
+	} else {
+		request.MaxParts, err = strconv.Atoi(query.Get("max-parts"))
+		if err != nil {
+			return
```
sfzeng: Could you please add a log for each error return?

s3/pkg/conf/tidb.sql
```diff
@@ -123,6 +123,26 @@ CREATE TABLE IF NOT EXISTS `multiparts` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+--
+-- Table structure for table `objectparts`
+--
+
+DROP TABLE IF EXISTS `objectparts`;
```
sfzeng: With this drop operation, every time restart the docker will remove those data already exist.

s3/pkg/service/multipart.go
```diff
@@ -238,16 +240,18 @@ func (s *s3Service) UploadPart(ctx context.Context, stream pb.S3_UploadPartStrea
 		return err
 	}
 
+	md5Writer := md5.New()
```
sfzeng: IMO, as ETag is calculated here now, then each backend driver no need to calculate it again, but just get the ETag from the backend storage if it provided, so the backend driver also need to do some change, right?

s3/pkg/service/multipart.go
```diff
@@ -257,7 +261,24 @@ func (s *s3Service) UploadPart(ctx context.Context, stream pb.S3_UploadPartStrea
 		log.Errorln("failed to upload part to backend. err:", err)
 		return err
 	}
-	uploadResponse.ETag = result.ETag
```
sfzeng: IMO, it's better to get ETag from the response, and check if it null, and it it is not null, check if it is the same as the one calculated here.

s3/pkg/service/multipart.go
```diff
@@ -306,6 +327,40 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
 		return err
 	}
 
+	md5Writer := md5.New()
+	var totalSize int64 = 0
+	for i := 0; i < len(in.CompleteParts); i++ {
+		if in.CompleteParts[i].PartNumber != int64(i+1) {
+			log.Infof("wrong order for part number. ")
```
sfzeng: Is it better to use log.Errorf here?

s3/pkg/service/multipart.go
```diff
@@ -306,6 +327,40 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
...
+			err = ErrInvalidPart
+			return err
+		}
+		part, ok := multipart.Parts[i+1]
+		if !ok {
+			log.Infoln("missed object part. partno:", i)
```
sfzeng: Same here

s3/pkg/service/multipart.go
```diff
@@ -306,6 +327,40 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
...
+			err = ErrInvalidPart
+			return err
+		}
+
+		if part.Etag != in.CompleteParts[i].ETag {
+			log.Infoln("part etag in meta store is not the same with client's part, partno:", i)
```
sfzeng: Same here.

s3/pkg/service/multipart.go
```diff
@@ -306,6 +327,40 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
...
+			return err
+		}
+		part, ok := multipart.Parts[i+1]
+		if !ok {
+			log.Infoln("missed object part. partno:", i)
+			err = ErrInvalidPart
+			return err
+		}
+
+		if part.Etag != in.CompleteParts[i].ETag {
```
sfzeng: Is it better to check the ETag in each UploadPart API instead of in CompleteMultipart API?

s3/pkg/service/multipart.go
```diff
@@ -446,6 +504,66 @@ func (s *s3Service) AbortMultipartUpload(ctx context.Context, in *pb.AbortMultip
...
+	uploadId := in.UploadId
+
+	multipart, err := s.MetaStorage.GetMultipart(bucketName, objectKey, uploadId)
+	if err != nil {
+		log.Errorln("failed to get multipart info. err:", err)
+		return err
+	}
+
+	isAdmin, tenantId, _, err := util.GetCredentialFromCtx(ctx)
+	if err != nil && isAdmin == false {
```
sfzeng: IOHM, no need to check isAdmin == false here, pleas check, thanks.

s3/pkg/service/multipart.go
```diff
@@ -446,6 +504,66 @@ func (s *s3Service) AbortMultipartUpload(ctx context.Context, in *pb.AbortMultip
...
+			})
+
+			if int64(len(out.Parts)) > in.MaxParts {
+				break
+			}
+		}
+	}
+	if int64(len(out.Parts)) == in.MaxParts+1 {
+		out.IsTruncated = true
+		out.NextPartNumberMarker = out.Parts[out.MaxParts].PartNumber
```
sfzeng: Does NextPartNumberMarker equals the PartNumber of current MaxPart?

api/pkg/s3/api-response.go
sunfch: thanks, will do it

api/pkg/s3/api-response.go
sunfch: @sfzeng ,  when parseListObjectPartsQuery is called,  the error log is printed at caller. Is it necessary to print log here?

s3/pkg/conf/tidb.sql
sunfch: thanks, will delete the line

s3/pkg/service/multipart.go
sunfch: I think we should no longer check the ETag from any backend, so it is not necessary to return ETag from backend.

s3/pkg/service/multipart.go
sunfch: @sfzeng , Thanks for your viewing,  I have passed ETag from client to backend, and the check of ETag should be done by backend.  because different backend calculate ETag with different ways,  the validity of data should be done at backend

s3/pkg/service/multipart.go
sunfch: thanks, will use Errorln instead of infof

s3/pkg/service/multipart.go
sunfch: thanks, will use Errorln instead of infof

s3/pkg/service/multipart.go
sunfch: thanks, will use Errorln instead of infof

s3/pkg/service/multipart.go
sunfch: CompleteParts[i].ETag is from client, so it is necessary to check if the ETag from user is correct.

s3/pkg/service/multipart.go
sunfch: thanks, will do it

s3/pkg/service/multipart.go
sunfch: thanks for your reviewing, yes, it is. if there are more than maxPart items, it indicate that IsTruncated  should be true,  and set NextPartNumberMarker as  the PartNumber of current MaxPart.

api/pkg/s3/api-response.go
sfzeng: I just thought that would be better for debug.

s3/pkg/service/multipart.go
sfzeng: But what if data damaged during transferring?

s3/pkg/service/multipart.go
sfzeng: But it sees when sd.UploadPart is called, ETag has not calculated yet, so the backend driver cannot get the ETag?

s3/pkg/service/multipart.go
sfzeng: OK

api/pkg/s3/api-response.go
sunfch: ok, will add log here, thanks

s3/pkg/service/multipart.go
sunfch: so I will add the etag check in s3 grpc server layer, and also pass etag to backend ?

PR #798, opened 2019-12-11
support migration through CopyObjectPart API privided by OSDS S3 service

sfzeng:
For those objects sith big size, we need to migrate it through the API of CopyObjectPart provided by OSDS S3 service.

**Release note**:

```release-note
Migration among AWS S3, Azure Blob, and Huawei OBS are test, other backends are left.
```


sfzeng:
This PR need to be seperated into two PRs.

PR #818, opened 2019-12-16
Support  the API of CopyObjectPart

### LINKED ISSUE #817: Support the API of CopyPart

> /kind feature

CopyPart API: Uploads a part by copying data from an existing object as data source. the requirement is from https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html. Precondition feature for CopyPart is supported in the PR

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:

Uploads a part by copying data from an existing object as data source when to put object with multipart.  You specify the data source by adding the request header x-amz-copy-source in your request and a byte range by adding the request header x-amz-copy-source-range in your request.  
You must initiate a multipart upload before you can upload any part. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #817 

**Release note**:

```release-note
Dont support to set acl in http header when put object with multipart now
```


s3/pkg/service/multipart.go
```diff
@@ -415,6 +415,7 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
 		Type:             ObjectTypeNormal,
 		Tier:             utils.Tier1,
 		Size:             totalSize,
+		Acl:              &pb.Acl{CannedAcl: "private"},
```
sfzeng: Hi @sunfch Do we support other ACL settings when CompleteMltipartUpload?

s3/pkg/service/multipart.go
```diff
@@ -567,3 +568,140 @@ func (s *s3Service) ListObjectParts(ctx context.Context, in *pb.ListObjectPartsR
...
+			break
+		default:
+			if srcObject.TenantId != tenantId {
+				err = ErrAccessDenied
+				return err
+			}
+		}
+	}
+
+	_, _, _, err = CheckRights(ctx, srcObject.TenantId)
```
sfzeng: As acl is already checked, do we need this CheckRights anymore?

s3/pkg/service/multipart.go
sunfch: @sfzeng  we still dont support to set acl with http header when to put object with multipart now, so      new object uploaded with multipart only has private acl.  thanks

s3/pkg/service/multipart.go
sunfch: thanks, will delete the code block

Issue #832, opened 2019-12-20
Missing encryption key when loading bucket SSE settings from DB


rhsakarpos:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:
The PR was closed and the code was re-written and a new PR raised.
In this, missed the code to load the encryption key from DB, to be used during a PUT to encrypt and to decrypt during a GET

**What you expected to happen**:
PUT object will upload encrypted object, GET object will allow download of decrypted object

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Gelato(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rhsakarpos:
fixed in #833

Issue #830, opened 2019-12-20
Unable to download contents of encrypted file from bucket


rhsakarpos:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:
Download a file from an encrypted bucket, the file's decrypted contents are seen in browser debugging, UI shows network unstable error

**What you expected to happen**:
The decrypted file to be downloaded

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
This at first look seems like a UI issue, needs further joint looking into

**Environment**:
- Gelato(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rhsakarpos:
@sfzeng please reopen in case needed, marking this closed

PR #833, opened 2019-12-24
fixes for #831 and #832 and #830

### LINKED ISSUE #831: Issue in loading configuration file for multi-cloud

>
/kind bug
> /kind feature

**What happened**:
The default conf. file is defined to be a file name, the code expects a folder name

To pick the values from the default conf. file when no other conf. file is provided

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rhsakarpos:
This PR has issue fixes for the SSE feature, these issues were caused in my analysis, since

- the older PR had to be closed and recreated
- non fcfs PR merges, leading to merge conflicts, sometimes with error-ed merge resolutions

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #831 #832 

**Release note**:

```release-note
```


rhsakarpos:
@sfzeng @kumarashit @nguptaopensds please review, thanks

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -113,17 +113,19 @@ func (t *TidbClient) GetBucket(ctx context.Context, bucketName string) (bucket *
...
+	tmp.ServerSideEncryption = &pb.ServerSideEncryption{}
 	sseOpts, sseErr := t.GetBucketSSE(ctx, tmp.Name)
 	if sseErr != nil {
 		return
 	}
 	sseType := "NONE"
 	if sseOpts != nil {
 		sseType = sseOpts.SseType
+		tmp.ServerSideEncryption.EncryptionKey = sseOpts.EncryptionKey
+		log.Infoln("encryption key is")
```
kumarashit: These two can be merged into one line

s3/pkg/service/object.go
```diff
@@ -375,13 +375,14 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
 			// decrypt and write
-			decErr, decBytes := utils.DecryptWithAES256(buf[0:n], bucket.ServerSideEncryption.EncryptionKey)
+			decErr, decBytes := utils.DecryptWithAES256(buf[0:left], bucket.ServerSideEncryption.EncryptionKey)
 			if decErr != nil {
 				log.Errorln("failed to decrypt data. err:", decErr)
 				return decErr
 			}
 			log.Infoln("successfully decrypted")
 			buf = decBytes
+			log.Infoln(string(decBytes))
```
kumarashit: Should this be logged? If yes, then please add message to in the log

s3/pkg/service/bucket.go
```diff
@@ -95,7 +95,6 @@ func (s *s3Service) CreateBucket(ctx context.Context, in *pb.Bucket, out *pb.Bas
 			return err
 		}
 	}
-	return nil
```
sfzeng: Why delete 'return nil' here?

s3/pkg/service/object.go
```diff
@@ -375,13 +375,14 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
 
 		if bucket.ServerSideEncryption.SseType == "SSE" {
 			// decrypt and write
-			decErr, decBytes := utils.DecryptWithAES256(buf[0:n], bucket.ServerSideEncryption.EncryptionKey)
+			decErr, decBytes := utils.DecryptWithAES256(buf[0:left], bucket.ServerSideEncryption.EncryptionKey)
```
sfzeng: Why use left instead of n here?

s3/pkg/meta/db/drivers/tidb/bucket.go
rhsakarpos: sheesh !!! missed removing debug logs, these are removed. thanks

s3/pkg/service/object.go
rhsakarpos: sheesh !!! missed removing debug logs, these are removed. thanks

s3/pkg/service/bucket.go
rhsakarpos: because the code after this became unreachable...

s3/pkg/service/object.go
rhsakarpos: i had used n, then some commit went in which does chunked download, and the variable changed to left. Decrypt stopped working, debugged and found this

s3/pkg/service/bucket.go
sfzeng: Ok, I missed the last return.

s3/pkg/service/object.go
sfzeng: But I cannot understand why this happened.  As my understanding, n bytes is saved in buf, not left bytes.

kumarashit: Will this try to decrypt whole object for every chunk? As we discussed, decryption works on whole object

rhsakarpos: for the decryption algorithm, we need the exact bytes, same as size of the object. n in this case will be chunk size, 2048, so decryption fails

rhsakarpos: @kumarashit yes, you are right. that is how it is working. what confused me here, is , in the S3 layer, we have this for loop, with the 'stream.send()' inside the loop. now, i figure this will send to the API layer (the caller), but there i do not see a loop

sfzeng: Have you tried object which size is more than 2M?

rhsakarpos: @sfzeng @kumarashit looks like the chunking will make some issue with SSE on large objects, thanks for the comment and discussons. I have now separated the code for sse and non-sse in GetObject. yet to test this, will do so, and add support for multi-part PUT and GET with SSE post DAITO, as planned earlier. Works ?

rhsakarpos: > Have you tried object which size is more than 2M?

yes @sfzeng, thanks. test screen shots below
encandnonencbuckets
enclargefile
nonenclargefile

rhsakarpos:
@kumarashit @sfzeng thanks for your review, have fixed the comments, please check

s3/pkg/service/object.go
```diff
@@ -375,7 +392,7 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
 
 		if bucket.ServerSideEncryption.SseType == "SSE" {
```
kumarashit: Didn't get this. Do we need this part of code, as it will be executed only if SSE is not set

rhsakarpos: correct, removed the unused code

s3/pkg/service/object.go
```diff
@@ -164,6 +164,7 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
 		limitedDataReader = data
 	}
 
+	actualSize := req.Size
 	// encrypt if needed
 	if bucket.ServerSideEncryption.SseType == "SSE" {
 		byteArr, _ := ioutil.ReadAll(limitedDataReader)
```
sfzeng: utils.EncryptWithAES256RandomKey may return error, so need to consider it before continue the next step.

s3/pkg/meta/db/drivers/tidb/object.go
```diff
@@ -34,12 +34,12 @@ func (t *TidbClient) GetObject(ctx context.Context, bucketName, objectName, vers
 	var row *sql.Row
 	if version == "" {
 		sqltext = "select bucketname,name,version,location,tenantid,userid,size,objectid,lastmodifiedtime,etag," +
-			"contenttype,customattributes,acl,nullversion,deletemarker,ssetype,encryptionkey,initializationvector,type,tier,storageMeta" +
+			"contenttype,customattributes,acl,nullversion,deletemarker,ssetype,encryptionkey,initializationvector,type,tier,storageMeta,encsize" +
```
sfzeng: It seems the schema of objects table in the database is changed, so the database initiation, in tidb.sql, also need to be changed accordingly.

rhsakarpos: fixed

s3/pkg/service/object.go
rhsakarpos: added check

sfzeng:
download_failed
Download failed, when SSE enabled, using ceph as the backend. Please check @rhsakarpos

sfzeng:
object
Here is the object.

kumarashit:
re-approved

PR #845, opened 2019-12-28
fix bugs that cross-cloud transition lifecycle rule does not works fo…

### LINKED ISSUE #844: Cross-Cloud transition rule does not works for big object.

>
> /kind bug

**What happened**:
1. Create cross-cloud transition rule for lifecycle management.
rule
2. Upload an object more than 5MB.
object
3. Wait lifecycle management schedule happens and check the object, it is not moved from one backend to another, and there are some error logs.
log

Object can be moved from one backend to another according to the lifecycle rule.

See above.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sfzeng:
…r big object

Fix bugs that cross-cloud transition lifecycle rule does not works for the big objects.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #844 

**Release note**:

```release-note
```

### DIFF
--- a/s3/pkg/datastore/azure/azure.go
+++ b/s3/pkg/datastore/azure/azure.go
@@ -146,34 +146,17 @@ func (ad *AzureAdapter) Get(ctx context.Context, object *pb.Object, start int64,
 	log.Infof("get object[Azure Blob], bucket:%s, objectId:%s\n", bucket, object.ObjectId)
 
 	blobURL := ad.containerURL.NewBlobURL(object.ObjectId)
-	log.Infof("blobURL:%v, size:%d\n", blobURL, object.Size)
 
-	var buf []byte
-	if end < object.Size - 1 {
-		count := end - start + 1
-		buf = make([]byte, count)
-		log.Debugf("start=%d, end=%d, count=%d\n", start, end, count)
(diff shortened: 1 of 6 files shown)


s3/pkg/datastore/azure/azure.go
```diff
@@ -146,34 +146,17 @@ func (ad *AzureAdapter) Get(ctx context.Context, object *pb.Object, start int64,
 	log.Infof("get object[Azure Blob], bucket:%s, objectId:%s\n", bucket, object.ObjectId)
 
 	blobURL := ad.containerURL.NewBlobURL(object.ObjectId)
-	log.Infof("blobURL:%v, size:%d\n", blobURL, object.Size)
-
-	var buf []byte
-	if end < object.Size - 1 {
```
kumarashit: Didn't get this part of the code, why are we removing this codition

sfzeng: As blobURL.Download can take care of all situations, small size and big size, no need to download to buffer first, and then return.

kumarashit: Ok

PR #853, opened 2019-12-31
PUT object versioning and List object versions

nguptaopensds:
This PR is for PUT object versioning on the s3 bucket and list object versions.

The PR is tested from postman and partly from dashboard.
The tests include:
1. Create backend AWS
2. Create bucket with versioning enabled
image
3. Upload object to versioned bucket
image
4. Upload object using postman 
image
5. List objects using postman
image
6. Verify the objects in metadata (DB)
image
7. Verify objects from AWS backend
image

**Release note**:

```release-note
```


s3/pkg/conf/tidb.sql
```diff
@@ -151,7 +151,7 @@ CREATE TABLE `objectparts` (
 CREATE TABLE IF NOT EXISTS `objects` (
   `bucketname` varchar(255) DEFAULT NULL,
   `name` varchar(255) DEFAULT NULL,
-  `version` bigint(20) UNSIGNED DEFAULT NULL,
```
kumarashit: Did you check if this change has any side effect on existing features

api/pkg/s3/bucketget.go
```diff
@@ -123,10 +123,19 @@ func parseListObjectsQuery(query url.Values) (request s3.ListObjectsRequest, err
 	return
 }
 
+func CreateListOfListObjectsResponse(bucketName string, request *s3.ListObjectsRequest,
+	listRsp *s3.ListObjectsResponse) (response []datatype.ListObjectsResponse) {
+	for _, o := range listRsp.ListOfListOfObjects {
```
kumarashit: The listRsp shld have a validatikn and logging

api/pkg/s3/bucketget.go
```diff
@@ -50,7 +50,7 @@ func (s *APIService) BucketGet(request *restful.Request, response *restful.Respo
 		return
 	}
 
-	rsp := CreateListObjectsResponse(bucketName, &req, lsRsp)
+	rsp := CreateListOfListObjectsResponse(bucketName, &req, lsRsp)
```
kumarashit: Bucket get will now only have  2d array?

s3/pkg/meta/db/drivers/tidb/object.go
```diff
@@ -88,9 +84,10 @@ func (t *TidbClient) GetObject(ctx context.Context, bucketName, objectName, vers
 		return
 	}
 	// TODO: getting multi-parts
-	timestamp := math.MaxUint64 - iversion
+	/*timestamp := math.MaxUint64 - iversion
 	timeData := []byte(strconv.FormatUint(timestamp, 10))
```
kumarashit: Remove this code if not applicable

kumarashit:
There are many code which are commented and seems part of old code that was put for versioning. Are they required?

s3/pkg/conf/tidb.sql
nguptaopensds: Yes, we tested it

api/pkg/s3/bucketget.go
nguptaopensds: Will add in testing phase, after RC4 during code coverage and functional testing

s3/pkg/meta/db/drivers/tidb/object.go
nguptaopensds: Okay will remove it

api/pkg/s3/bucketget.go
kumarashit: RC4 is the final one. This is the final release cut :)

sfzeng:
IMHO, no matter versioning is enabled or not, OSDS needs to keep the object key not changed. For example, when the user puts an object obj1, the object key should always be obj1, not obj1_xxxxxxxx.

kumarashit:
> 
> 
> IMHO, no matter versioning is enabled or not, OSDS needs to keep the object key not changed. For example, when the user puts an object obj1, the object key should always be obj1, not obj1_xxxxxxxx.

Does OSDS keep Object Key same as object name? Or is it Unique ID?

nguptaopensds:
> IMHO, no matter versioning is enabled or not, OSDS needs to keep the object key not changed. For example, when the user puts an object obj1, the object key should always be obj1, not obj1_xxxxxxxx.

This is done to keep the object key unique in OpenSDS and the cloud backend, since we are NOT enabling versioning on the cloud bucket...do share specific concerns, we can discuss, thanks

s3/pkg/utils/utils.go
```diff
@@ -139,3 +142,13 @@ func GetBackend(ctx context.Context, backedClient backend.BackendService, backen
...
 	return backend, nil
 }
+
+func GetVersionId() (versionId string, timeStamp int64) {
+	now := time.Now().UTC().Unix()
+	version := strconv.FormatUint(math.MaxUint64 - uint64(now),10)
+	return version, now
+}
+
+func GetSingleVersionId() (versionId string) {
```
kumarashit: Why this function?

s3/pkg/service/object.go
```diff
@@ -116,17 +116,159 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
 	}
 
-	req := &pb.PutObjectRequest{}
-	err = in.RecvMsg(req)
+	log.Infof("*********bucket:%s,key:%s,size:%d\n", req.BucketName, req.ObjectKey, req.Size)
+	bucket, err := s.MetaStorage.GetBucket(ctx, req.BucketName, true)
 	if err != nil {
-		log.Errorln("failed to get msg with err:", err)
-		return ErrInternalError
+		log.Errorln("get bucket failed with err:", err)
```
kumarashit: Ideally these code should be moved into another function and that function should be called by both GetVersionedObject and GetObject

s3/pkg/service/object.go
```diff
@@ -116,17 +116,159 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+	isBucketVersioned := bucket.Versioning.Status == utils.VersioningEnabled
+	if isBucketVersioned {
+		log.Info("Inside service/object.go PUT bucket version enabled")
+		versionId, now := utils.GetVersionId()
+		obj.VersionId = versionId
+		obj.LastModified = now
+		// update the cloud backend object name here
+		obj.ObjectKey = obj.ObjectKey + "_" + obj.VersionId
+	} else {
+		obj.VersionId = utils.GetSingleVersionId()
```
kumarashit: Why can't we simply put "" rather than doing a function call?

s3/pkg/service/object.go
```diff
@@ -116,17 +116,159 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+	ctx = context.WithValue(ctx, dscommon.CONTEXT_KEY_SIZE, req.Size)
+	ctx = context.WithValue(ctx, dscommon.CONTEXT_KEY_MD5, bodyMd5)
+	sd, err := driver.CreateStorageDriver(backend.Type, backend)
+	if err != nil {
+		log.Errorln("failed to create storage. err:", err)
+		return err
+	}
+	obj := &pb.Object{BucketName: req.BucketName, ObjectKey: req.ObjectKey}
+	isBucketVersioned := bucket.Versioning.Status == utils.VersioningEnabled
+	if isBucketVersioned {
```
kumarashit: Infact what i think is that we can simply move line# 203 to 211 in a separate function and call it from inside getObject().

sfzeng:
> > IMHO, no matter versioning is enabled or not, OSDS needs to keep the object key not changed. For example, when the user puts an object obj1, the object key should always be obj1, not obj1_xxxxxxxx.
> 
> This is done to keep the object key unique in OpenSDS and the cloud backend, since we are NOT enabling versioning on the cloud bucket...do share specific concerns, we can discuss, thanks

We have objectkey and objectid, objectkey is used by the end user, while objectid is the real key used in backend.

kumarashit:
> 
> 
> > > IMHO, no matter versioning is enabled or not, OSDS needs to keep the object key not changed. For example, when the user puts an object obj1, the object key should always be obj1, not obj1_xxxxxxxx.
> > 
> > 
> > This is done to keep the object key unique in OpenSDS and the cloud backend, since we are NOT enabling versioning on the cloud bucket...do share specific concerns, we can discuss, thanks
> 
> We have objectkey and objectid, objectkey is used by the end user, while objectid is the real key used in backend.

Then we can have ObjectKey same as object name and ObjectId can be unique. @nguptaopensds ??

s3/pkg/service/object.go
```diff
@@ -116,17 +116,159 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+	if bucket.Versioning.Status == utils.VersioningEnabled {
+		log.Info("Inside PUT Versioned Objects")
+		return PutVersionedObject(ctx, req, in, s)
+	} else {
+		log.Info("Inside PUT  Objects")
+		return PutObject(ctx, req, in, s)
+	}
+}
+
+func PutVersionedObject(ctx context.Context, req *pb.PutObjectRequest, in pb.S3_PutObjectStream, s *s3Service) error {
```
sfzeng: Most part of PutVersionedObject and PutObject are the same, please move those part into a function, and let  PutVersionedObject and PutObject call it.

s3/pkg/service/object.go
```diff
@@ -116,17 +116,159 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+		}
+	}
+	data := &StreamReader{in: in}
+	var limitedDataReader io.Reader
+	if req.Size > 0 { // request.ContentLength is -1 if length is unknown
+		limitedDataReader = io.LimitReader(data, req.Size)
+	} else {
+		limitedDataReader = data
+	}
+
```
sfzeng: Encryption is lost here.

s3/pkg/service/object.go
```diff
@@ -116,17 +116,159 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
...
+	}
+	obj := &pb.Object{BucketName: req.BucketName, ObjectKey: req.ObjectKey}
+	isBucketVersioned := bucket.Versioning.Status == utils.VersioningEnabled
+	if isBucketVersioned {
+		log.Info("Inside service/object.go PUT bucket version enabled")
+		versionId, now := utils.GetVersionId()
+		obj.VersionId = versionId
+		obj.LastModified = now
+		// update the cloud backend object name here
+		obj.ObjectKey = obj.ObjectKey + "_" + obj.VersionId
```
sfzeng: VersionId should not be a part of ObjectKey.

s3/pkg/service/object.go
```diff
@@ -221,6 +363,7 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
 	obj.Size = req.Size
 	obj.Location = backendName
 
+
```
sfzeng: Format code please.

s3/pkg/service/object.go
```diff
@@ -982,9 +1127,42 @@ func (s *s3Service) ListObjects(ctx context.Context, in *pb.ListObjectsRequest,
...
+			}
+			objList.Objects = append(objList.Objects, &object)
+			mapNameToListOfObjects[object.ObjectKey] = objList
+		} else {
+			objList.Objects = append(objList.Objects, &object)
+			mapNameToListOfObjects[object.ObjectKey] = objList
+		}
+	}
+
+	out.ListOfListOfObjects = make([]*pb.ListObjects, len(mapNameToListOfObjects))
```
sfzeng: Does this be compatible with AWS S3, otherwise many s3 client tools cannot be used.

api/pkg/s3/bucketget.go
nguptaopensds: yes

s3/pkg/service/object.go
nguptaopensds: We'll do it in next release

s3/pkg/service/object.go
nguptaopensds: Will do it after this release

s3/pkg/service/object.go
nguptaopensds: will add in new code

s3/pkg/service/object.go
nguptaopensds: Will do it after Daito

s3/pkg/service/object.go
nguptaopensds: It's to get the new object key after we enable versioning

s3/pkg/service/object.go
nguptaopensds: We intend to use this function in future for getting single version id

s3/pkg/service/object.go
nguptaopensds: ok

s3/pkg/utils/utils.go
nguptaopensds: replied earlier

PR #855, opened 2020-01-02
GET Object with version ID

nguptaopensds:
**What this PR does / why we need it**:
This PR is for Getting the object with specific version ID

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:
Request URL
```http://<host_ip>:8089/v1/s3/<bucket_name>/<object_name>?versionId=<version_id>```

Tested with postman:
image

**Release note**:

```release-note
```

PR #856, opened 2020-01-02
Delete Object with specific version ID

nguptaopensds:
This PR is for deleting the object with specific version ID

DELETE Request URL
```http://<host_ip>:8089/v1/s3/<bucket_name>/<object_name>?versionId=<version_id>```

Tested using postman
image

**Release note**:

```release-note
```


nguptaopensds:
The CI is failing because GET Object PR has some changes which are used in this PR, so if PR #855 is merged , the CI will pass

sfzeng:
Hi @nguptaopensds , as we discussed on slack before that currently, we have to be AWS S3 compatible, please check https://docs.aws.amazon.com/AmazonS3/latest/dev/ObjectVersioning.html. The current implementation of delete versioning object is not compatible with it.

PR #860, opened 2020-01-03
Bug fix #840

### LINKED ISSUE #840: Versioning settings is not deleted after related bucket is deleted.

>
> /kind bug

**What happened**:
After a bucket is deleted, it's related versioning setting is left in database.
Here are the existing buckets:
image
Here are the versioning settings in database:
image
And this will cause error when I created a new bucket which name is the same as what has been deleted:
image

No versioning setting left after the bucket is deleted.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nguptaopensds:
Bug fix for #840 , Delete bucket doesn't delete entries from bucket_versionopts table.

#840 

**Release note**:

```release-note
```


s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -483,6 +483,12 @@ func (t *TidbClient) DeleteBucket(ctx context.Context, bucket *Bucket) error {
 	_, err := t.Client.Exec(sqltext, bucket.Name)
 	if err != nil {
 		return handleDBError(err)
+	} else {
+		sqltext := "delete from bucket_versionopts where bucketname=?;"
```
sfzeng: Suggest to use transaction here, make sure buckets and bucket_versionopts both success, otherwise roll back.

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -483,6 +483,12 @@ func (t *TidbClient) DeleteBucket(ctx context.Context, bucket *Bucket) error {
...
+		_, err := t.Client.Exec(sqltext, bucket.Name)
+		if err != nil {
```
kumarashit: If bucket has no versioning set, If there is no entry in the bucket_versionopts for the bucket then this will throw error. This is not correct

nguptaopensds: For all buckets, whether versioned or not, bucket_versionopts will have an entry with status enabled/disabled , so this won't be a problem

s3/pkg/meta/db/drivers/tidb/bucket.go
nguptaopensds: yes did the code change, will push new commit

sfzeng: This is not transaction, see DeleteObject in tidb/object.go please.

skdwriting: @nguptaopensds , this comment is handled?

nguptaopensds:
Not required yet

PR #874, opened 2020-01-07
in yig storage backend, when put object, break the loop of read data …

### LINKED ISSUE #875: The PostObject with yig storage backend may fail to save object data due to the hang of read operation.
This is a bug, and it is only related to yig storage backend when perform post object.
If yig backend reads data and it meets the eof but the returned data size is not 0, the yig backend will hang there because it will try to read more data from the upper reader. 
>
> /kind bug
> /kind feature

**What happened**:
When perform PostObject with yig storage backend, yig may fail to save the object data.
yig backend can save the object data successfully when perform PostObject operation.

PostObject with 1024 data size and use yig storage backend may reproduce this issue.

it is only related to yig storage backend when perform post object.
If yig backend reads data and it meets the eof but the returned data size is not 0, the yig backend will hang there because it will try to read more data from the upper reader. 

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hopkings2008:
…if eof flag is set.

This PR fixes the problem that yig storage backend may fail to save object data through PostObject interface.
If yig backend gets the EOF from reader but gets the non-zero data size, yig backend will try to read data again, and it will hang there because grpc server doesn't have more data to send.
This PR sets the EOF flag when reads data from grpc server and break the next read loop if EOF flag is set.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #875 

**Release note**:

```release-note
```


sfzeng:
Hi @hopkings2008 could you please explain what is the scenario that yig backend gets the EOF from reader but gets the non-zero data size, I mean when will this happend? Is it always happen for the last read?

sfzeng:
> > Hi @hopkings2008 could you please explain what is the scenario that yig backend gets the EOF from reader but gets the non-zero data size, I mean when will this happend? Is it always happen for the last read?
> 
> @sfzeng the yig backend storage driver will use the io reader provided by grpc server to read data from client and write them into ceph cluster. Because io reader has a buffer in it, the loop time to read data from yig backend is the one whole data size divided by buffer size, and if the last one is less than the buffer size, the reader will return eof and the remaining data to the yig backend.

Thanks for your explain, but why it only affect PostObject, not affect Putobject?

hopkings2008:
> > > Hi @hopkings2008 could you please explain what is the scenario that yig backend gets the EOF from reader but gets the non-zero data size, I mean when will this happend? Is it always happen for the last read?
> > 
> > 
> > @sfzeng the yig backend storage driver will use the io reader provided by grpc server to read data from client and write them into ceph cluster. Because io reader has a buffer in it, the loop time to read data from yig backend is the one whole data size divided by buffer size, and if the last one is less than the buffer size, the reader will return eof and the remaining data to the yig backend.
> 
> Thanks for your explain, but why it only affect PostObject, not affect Putobject?

PutObject contains the size of the object body, but postObject request doesn't contain the separate size of object body. So, it only affects the post object operation.

Issue #869, opened 2020-01-05
Download failed when using ceph as the backend and enable SSE.


sfzeng:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug

**What happened**:
1. Creat a bucket using ceph as it's backend, and enable SSE.
![ceph_bucket_sse](https://user-images.githubusercontent.com/38932432/71779532-00c93980-2ff2-11ea-89f0-6e9a850c6a5f.png)
2. Upload an object:
![ceph_sse_upload](https://user-images.githubusercontent.com/38932432/71779582-29513380-2ff2-11ea-8727-7c42ea599a59.png)
3. Download it failed:
![ceph_sse_download_failed](https://user-images.githubusercontent.com/38932432/71779613-4554d500-2ff2-11ea-93e3-04f45c54a652.png)
4. Check log:
![ceph_sse_download_failed_log](https://user-images.githubusercontent.com/38932432/71779616-556cb480-2ff2-11ea-99ee-d612ef11beed.png)

**What you expected to happen**:
Download it successfully.

**How to reproduce it (as minimally and precisely as possible)**:
See the above steps.

**Anything else we need to know?**:

**Environment**:
- Gelato(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rhsakarpos:
@sfzeng please reopen in case needed, marking this closed

PR #876, opened 2020-01-07
fix fo r#869

rhsakarpos:
fix for #869 

#869 
image

**Release note**:

```
This also adds a fix to delete SSE settings on bucket delete
```

### DIFF
--- a/s3/pkg/meta/db/drivers/tidb/bucket.go
+++ b/s3/pkg/meta/db/drivers/tidb/bucket.go
@@ -478,13 +478,37 @@ func (t *TidbClient) CountObjects(ctx context.Context, bucketName, prefix string
 	return &rsp, err
 }
 
-func (t *TidbClient) DeleteBucket(ctx context.Context, bucket *Bucket) error {
+func (t *TidbClient) DeleteBucket(ctx context.Context, bucket *Bucket) (err error) {
+	var sqlTx *sql.Tx
+	var tx interface{}
+	tx, err = t.Client.Begin()
+
+	defer func() {
+		if err == nil {
+			err = sqlTx.Commit()
+		}
+		if err != nil {
+			sqlTx.Rollback()
+		}
+	}()
+
+	sqlTx, _ = tx.(*sql.Tx)
+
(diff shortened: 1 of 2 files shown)


rhsakarpos:
@kumarashit @sfzeng @nguptaopensds @anvithks please review, thanks

s3/pkg/service/object.go
```diff
@@ -410,6 +427,7 @@ func GetObject(ctx context.Context, req *pb.GetObjectInput, stream pb.S3_GetObje
 }
 
 func GetEncObject(ctx context.Context, req *pb.GetObjectInput, stream pb.S3_GetObjectStream, s *s3Service) error {
+
```
sfzeng: Format code please

rhsakarpos: did that, ran go fmt multiple times, still looks like this...am i missing something ?
image

sfzeng: How did you run it?

rhsakarpos: like this...

s3/pkg/meta/db/drivers/tidb/bucket.go
```diff
@@ -478,13 +478,37 @@ func (t *TidbClient) CountObjects(ctx context.Context, bucketName, prefix string
...
+	}()
+
+	sqlTx, _ = tx.(*sql.Tx)
+
 	sqltext := "delete from buckets where bucketname=?;"
-	_, err := t.Client.Exec(sqltext, bucket.Name)
+	_, err = sqlTx.Exec(sqltext, bucket.Name)
 	if err != nil {
-		return handleDBError(err)
+		err = handleDBError(err)
```
sfzeng: Don't we need to return here?

rhsakarpos: to err on the side of caution (based on sample code i saw on a blog), did not return...yes, that might be more efficient in the sense we will not try the second SQL, but even with current code, the intent is surely met, either BOTH SQL are executed or NONE are executed. Tested this. Thanks

PR #886, opened 2020-01-31
Versioning feature support for multi-cloud

nguptaopensds:
This PR is for versioning support in multi-cloud. 

The testing report yet to be added

**Release note**:

```release-note
```


api/pkg/s3/bucketget.go
```diff
@@ -181,6 +208,72 @@ func CreateListObjectsResponse(bucketName string, request *s3.ListObjectsRequest
 	return
 }
 
+// this function refers to GenerateListObjectsResponse in api-response.go from Minio Cloud Storage.
```
kumarashit: What is this comment

api/pkg/s3/objectdelete.go
```diff
@@ -28,7 +28,12 @@ func (s *APIService) ObjectDelete(request *restful.Request, response *restful.Re
 	url := request.Request.URL
 	bucketName := request.PathParameter("bucketName")
 	objectName := request.PathParameter("objectKey")
-	version := url.Query().Get("versionId")
+	versionId := request.Request.URL.Query().Get(common.REQUEST_QUERY_VERSION_ID)
+	if versionId != ""{
+		objectName = objectName + "_" + versionId
```
sfzeng: Does this mean, for a specific object, if versioning is enabled, then it's key will be different from the one with versioning disabled?

api/pkg/s3/routbucketget.go
```diff
@@ -27,8 +27,8 @@ func (s *APIService) RouteBucketGet(request *restful.Request, response *restful.
 		s.BucketAclGet(request, response)
 	} else if IsQuery(request, "uploads") {
 		s.ListBucketUploadRecords(request, response)
-	} else if IsQuery(request, "versioning") {
-		//TODO
+	} else if IsQuery(request, "allversions") {
```
sfzeng: Is this be compatible with aws s3? From https://docs.aws.amazon.com/AmazonS3/latest/dev/list-obj-version-enabled-bucket.html, the URL of listing all object versions in a bucket is supposed to be GET /?versions HTTP/1.1, while Retrieving All Versions of Only a Specific Object is supposed to be GET /?versions&prefix=objectName HTTP/1.1.

sfzeng:
Hi @nguptaopensds , could you please explain what will the end user get in the following sceanrios:
1. Getting object without version id specifed if versioning is enabled?
2. Deleting object without version id specifed if versioning is enabled?
3. Get bucket without parameters in the URL?
4. How many versioning related APIs are compatible with AWS S3, how many not?

PR #888, opened 2020-02-03
fix the bug that downloading object failed

### LINKED ISSUE #887: Download object failed.

>
> /kind bug

**What happened**:
Upload a file which only contains an 'abc' string, then download it, download runs failed,  and got error log as below:
download_failed_20200203

The file can be download successfully.

Follow the steps described above.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sfzeng:
This PR is to fix the bug that downloading object failed.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #887 

The root cause of this bug is the way to calculate timeout is not correct, if object size is less than 5Bytes, then the timeout will be 0, that is not right. So, in this PR, the way to calculate timeout is modified.

**Release note**:

```release-note
```

### DIFF
--- a/api/pkg/utils/utils.go
+++ b/api/pkg/utils/utils.go
@@ -17,12 +17,16 @@ package utils
 import (
 	"encoding/json"
 	"errors"
+	"math"
 	"math/rand"
 	"os"
 	"reflect"
+	"strconv"
 
 	log "github.com/sirupsen/logrus"
 )
+
+const DEFAULT_TIMEOUT_SEC int64 = 5 // Same as the default timeout of go-micro.
 
 //remove redundant elements
 func RvRepElement(arr []string) []string {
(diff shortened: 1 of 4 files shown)


api/pkg/utils/utils.go
```diff
@@ -17,13 +17,17 @@ package utils
...
+	"math"
 	"math/rand"
 	"os"
 	"reflect"
+	"strconv"
 
 	log "github.com/sirupsen/logrus"
 )
 
+var DEFAULT_TIMEOUT_SEC int64 = 5 // Same as the default timeout of go-micro.
```
NajmudheenCT: How about making this as a const, as we dont expect runtime change of this value.
const DEFAULT_TIMEOUT_SEC [int64 ] = 5

sfzeng: Thanks, done.

PR #891, opened 2020-02-08
fix for #889

### LINKED ISSUE #889: Object is corrupted with sse enabled.

>
> /kind bug

**What happened**:
1. Create a bucket with sse enabled.
2. Upload a *.exe file which size is about 10MB.
3. Download it and open it failed, got the error message as below:
image (1)

1. Open the file successfully as below:
image (2)

Follow the steps described above.

According to my test, it seems exe files with size less than 5MB works fine, so i think the bug is about multipart upload with sse enabled.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rhsakarpos:
This fixes #889 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #889 

**Release note**:

```
- this PR adds code to decrypt in 5 MB chunks (just like the encrypt for multipart is done in chunks of 5 MB). So, 

- download the whole file from cloud backend
- decrypt in 5 MB chunks
- re-assemble and send to browser

Have tested with .exe files and .txt files, lesser than 5MB and larger than 10 MB, did a byte comparison of original file and downloaded file. Screenshots shared below

![image](https://user-images.githubusercontent.com/48434725/74078957-c1f62b80-4a56-11ea-8929-25cae654a120.png)

![image](https://user-images.githubusercontent.com/48434725/74078965-dfc39080-4a56-11ea-987e-2560fb30fcaa.png)

![image](https://user-images.githubusercontent.com/48434725/74078970-efdb7000-4a56-11ea-96fa-bcad4d8ec8ce.png)

```

### DIFF
--- a/s3/pkg/service/multipart.go
+++ b/s3/pkg/service/multipart.go
@@ -17,6 +17,8 @@ package service
 import (
 	"bytes"
 	"context"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/md5"
 	"encoding/hex"
 	"io"
@@ -261,23 +263,27 @@ func (s *s3Service) UploadPart(ctx context.Context, stream pb.S3_UploadPartStrea
 	md5Writer := md5.New()
 	data := &StreamReader{in: stream}
 	limitedDataReader := io.LimitReader(data, size)
+
 	var dataReader io.Reader
 	// encrypt if needed
 	if bucket.ServerSideEncryption.SseType == "SSE" {
 
-		var readerErr error
-		tempReader, readerErr := utils.WrapEncryptionReader(limitedDataReader,
-			bucket.ServerSideEncryption.EncryptionKey,
-			bucket.ServerSideEncryption.InitilizationVector)
-		if readerErr != nil {
-			log.Errorln("failed to get encrypted reader with err:", readerErr)
-			return readerErr
-		}
-		buf := new(bytes.Buffer)
-		buf.ReadFrom(tempReader)
-		size = int64(buf.Len())
+		var out bytes.Buffer
 
-		dataReader = io.TeeReader(io.LimitReader(bytes.NewReader(buf.Bytes()), size), md5Writer)
+		block, err := aes.NewCipher(bucket.ServerSideEncryption.EncryptionKey)
+		if err != nil {
+			panic(err)
+		}
+		stream := cipher.NewCTR(block, bucket.ServerSideEncryption.InitilizationVector)
+		writer := &cipher.StreamWriter{S: stream, W: &out}
+		// Copy the input to the output buffer, encrypting as we go.
+		if _, err := io.Copy(writer, limitedDataReader); err != nil {
+			panic(err)
+		}
+
+		log.Infof("length of encrypted bytes-%v", len(out.Bytes()))
+		dataReader = io.TeeReader(bytes.NewReader(out.Bytes()), md5Writer)
+
 	} else {
 		dataReader = io.TeeReader(limitedDataReader, md5Writer)
 	}
@@ -296,6 +302,7 @@ func (s *s3Service) UploadPart(ctx context.Context, stream pb.S3_UploadPartStrea
 		UploadId: uploadId,
 		ObjectId: multipart.ObjectId},
 		int64(partId), size)
+
 	if err != nil {
 		log.Errorln("failed to upload part to backend. err:", err)
 		return err
@@ -461,6 +468,7 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
 		Type:             ObjectTypeNormal,
 		Tier:             multipart.Metadata.Tier,
 		Size:             totalSize,
+		EncSize:          totalSize,
 		Location:         multipart.Metadata.Location,
 		Acl:              &multipart.Metadata.Acl,
 	}
--- a/s3/pkg/service/object.go
+++ b/s3/pkg/service/object.go
@@ -2,6 +2,7 @@
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
+// you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
@@ -15,8 +16,12 @@
 package service
 
 import (
+	"bytes"
 	"context"
+	"crypto/aes"
+	"crypto/cipher"
 	"io"
+	"io/ioutil"
 	"net/url"
 	"time"
 
@@ -35,8 +40,6 @@ import (
 	pb "github.com/opensds/multi-cloud/s3/proto"
 	log "github.com/sirupsen/logrus"
 )
-
-var ChunkSize int = 2048
 
 func (s *s3Service) CreateObject(ctx context.Context, in *pb.Object, out *pb.BaseResponse) error {
 	log.Infoln("CreateObject is called in s3 service.")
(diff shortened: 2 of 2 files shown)


rhsakarpos:
image

image

image

@sfzeng @kumarashit please review, thanks

s3/pkg/service/object.go
```diff
@@ -2,6 +2,7 @@
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
+// you may not use this file except in compliance with the License.
```
sfzeng: Why add this?

s3/pkg/service/object.go
```diff
@@ -355,45 +358,77 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
 		return err
 	}
 
-	var finalReader io.Reader
-	finalReader = reader
+	allBytes, _ := ioutil.ReadAll(reader)
```
sfzeng: Will this read all the data from the reader? For example, if the object is 2G, will this read all 2G bytes to allBytes? If it is like this, this will consume a lot of memory.

s3/pkg/service/object.go
```diff
@@ -355,45 +358,77 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
+	log.Infof("allBytes.len - %v", len(allBytes))
+
+	allDecBytes := make([]byte, 0)
+
 	if bucket.ServerSideEncryption.SseType == "SSE" {
-		finalReader, _ = utils.WrapAlignedEncryptionReader(reader, 0,
-			bucket.ServerSideEncryption.EncryptionKey,
-			bucket.ServerSideEncryption.InitilizationVector)
-	}
+		// this should match the upload chunk size during multipart upload, which is 5MB (5*1024*1024). During multipart
```
sfzeng: 5MB is not specific, it's depends on client, dashboard use 5MB as default.

s3/pkg/service/object.go
```diff
@@ -355,45 +358,77 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
 		}
-		// From https://golang.org/pkg/io/, there is the following statement.
-		// Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) ==
-		// 0. Callers should treat a return of 0 and nil as indicating that nothing happened; in particular it does not indicate EOF.
-		// If n is equal 0, it indicate that there is no more data to read
-		if n == 0 {
-			log.Infoln("reader return zero bytes.")
-			break
+
+		decChunkBytes := make([]byte, 0)
```
sfzeng: Why need decChunkBytes here, is it possible to use allBytes directly?

s3/pkg/service/object.go
rhsakarpos: yes, it will read all, since after the read, we need to decrypt in chunks...there is no multipart download from cloud...checked...please suggest any alternative, thanks

s3/pkg/service/object.go
rhsakarpos: yes, that is why used 5 MB here. One thing i could think of was, to update the API for ```initMultipart``` to take in the chunk size as an input, persist it and use it here

s3/pkg/service/object.go
rhsakarpos: will remove

s3/pkg/service/object.go
rhsakarpos: to decrypt in 5MB chunks...so decrypt each chunk of 5MB, keep appending the decrypted byte[ ] to allBytes, till fully done

s3/pkg/service/object.go
sfzeng: Is it necessary to change the encryption chunk according to the part size? How about using a fixed chunk size independent of it?

s3/pkg/service/multipart.go
```diff
@@ -461,6 +468,7 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
 		Type:             ObjectTypeNormal,
 		Tier:             multipart.Metadata.Tier,
 		Size:             totalSize,
+		EncSize:          totalSize,
```
sfzeng: Hi @rhsakarpos , could you please explain what EncSize means, if it is the same as Size, is it necessary?

s3/pkg/service/object.go
sfzeng: There is download range for cloud storage, I think. If the size is big, then download all will use too much memory.

s3/pkg/service/object.go
```diff
@@ -355,45 +357,77 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
+
+		for i := 0; i < modVal; i++ {
+			decThouBytes := make([]byte, 0)
+			for j := (i) * chunkSize; j < (i+1)*chunkSize; j++ {
+				decThouBytes = append(decThouBytes, allBytes[j])
+			}
+
+			// encrypt in chunks of chunkSize, write file to disk
+			var out bytes.Buffer
+			block, err := aes.NewCipher(bucket.ServerSideEncryption.EncryptionKey)
```
sfzeng: Hi @rhsakarpos , here use sse configuration of bucket, then what if a user enable sse of a bucket first, and upload an object, then disable the sse of the bucket, and download the object?

s3/pkg/service/object.go
rhsakarpos: i checked AWS docs, could not find any...download with range...

s3/pkg/service/object.go
rhsakarpos: thought of that, but ruled it out, since, it will get messy during a multi part upload...example, if we encrypt each 2 MB chunk, and multipart chunks each 5MB, it is a problem

s3/pkg/service/multipart.go
rhsakarpos: typically, depending on the algorithm used, the encryption on a byte[ ] changes the size of the byte[ ]. This size is needed when we decrypt the encrypted byte[ ]

s3/pkg/service/object.go
rhsakarpos: that will fail. currently, we do not have disable SSE on bucket for this reason.

s3/pkg/service/object.go
```diff
@@ -355,45 +357,77 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
...
-		err = stream.Send(&pb.GetObjectResponse{ErrorCode: int32(ErrNoErr), Data: buf[0:n]})
+		// decrypt in chunks of 5MB
+		var out bytes.Buffer
+		block, err := aes.NewCipher(bucket.ServerSideEncryption.EncryptionKey)
 		if err != nil {
-			log.Infof("stream send error: %v\n", err)
-			break
+			panic(err)
+		}
+		stream := cipher.NewCTR(block, bucket.ServerSideEncryption.InitilizationVector)
```
sunfch: Where is the variable "InitilizationVector" initialized？

s3/pkg/service/object.go
```diff
@@ -355,45 +357,77 @@ func (s *s3Service) GetObject(ctx context.Context, req *pb.GetObjectInput, strea
 		return err
 	}
 
-	var finalReader io.Reader
-	finalReader = reader
+	allBytes, _ := ioutil.ReadAll(reader)
+
+	log.Infof("allBytes.len - %v", len(allBytes))
+
+	allDecBytes := make([]byte, 0)
```
sunfch: How to handle with range download when start offset is not aligned?

s3/pkg/service/object.go
sfzeng: https://docs.aws.amazon.com/AmazonS3/latest/dev/bucket-encryption.html, here, search range please.

s3/pkg/service/object.go
rhsakarpos: in our GET Object API, we do not have or support start offset based GET. So, not an issue, in my understanding

s3/pkg/service/object.go
rhsakarpos: thanks, did not find...am i missing something ?
image

sfzeng: Hi @rhsakarpos , have you found the range definition now? Maybe google with key word aws s3, get range can help.

s3/pkg/service/multipart.go
sfzeng: OK, Eh.... but why EncSize equals Size?

s3/pkg/service/object.go
sfzeng: I think user can specify range in header to get a specific range, not need new API.

rhsakarpos: Ok

s3/pkg/service/object.go
rhsakarpos: yes, found this...correct ?
```
GET /example-object HTTP/1.1
            Host: example-bucket.s3.<Region>.amazonaws.com
            x-amz-date: Fri, 28 Jan 2011 21:32:02 GMT
            Range: bytes=0-9
            Authorization: AWS AKIAIOSFODNN7EXAMPLE:Yxg83MZaEgh3OZ3l0rLo5RTX11o=
            Sample Response with Specified Range of the Object Bytes
```

s3/pkg/service/multipart.go
rhsakarpos: in this streaming encryption algorithm, the size is the same

rhsakarpos:
the code has gone out of sync with current code, been over a month, closing this

PR #894, opened 2020-03-04
Separate S3 APIs from other multi-cloud APIs and change the authentification of S3 APIs to be compatible with AWS S3.

### LINKED ISSUE #893: S3 APIs are not AWS S3 compatible.

>
> /kind bug

**What happened**:
Two parts:
1. Currently, the URL of S3 APIs is like http(s)://ip:port/v1/s3/bucketname, which is not compatible with AWS S3, v1/s3/ should be removed.
At the same time, the authentification used a private algorithm, which is not compatible with AWS S3 too.
2. Each s3 API did not do AWS s3 compatilbe test, they may not be AWS S3 compatible.

Because of the above reasons, the third party s3 client, like the s3 browser, cannot be used to test multi-cloud.

The third-party s3 client, like the s3 browser, cannot be used to test multi-cloud.

NA

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sfzeng:
This PR is the first PR to make multi-cloud be compatible with AWS S3. Two things include in this PR:
1. Separate S3 APIs from other multi-cloud APIs.
2. Change the authentification of S3 APIs, make it support auth v2/v4, which is the same as AWS S3.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #893 

1. This is the first PR for AWS S3 compatible issue, there may have some duplicate code between s3api and api. In the next PR I will remove those duplicate code by creating some public folders to store those duplicate code.
2. This PR just fix the URL and authentification issues, but not make sure all S3 APIs be compatible with AWS S3. We will do AWS S3 compatible test next, and raise new PRs if needed.

**Release note**:

```
As S3 APIs are separated from other APIs, the dashboard needs to change accordingly. The dashboard will be changed later.
```


docker-compose.yml
```diff
@@ -38,6 +38,24 @@ services:
...
+      - 8090:8090
+    environment:
+      - MICRO_SERVER_ADDRESS=:8090
+      - MICRO_REGISTRY=mdns
+      - OS_AUTH_AUTHSTRATEGY=keystone
+      - OS_AUTH_URL=http://10.10.3.151/identity
+      - OS_USER_DOMIN_ID=Default
+      - OS_USERNAME=opensds
+      - OS_PASSWORD=opensds@123
       - OS_PROJECT_NAME=service
```
sunfch: missing  OS_PROJECT_NAME in s3api ?

s3api/cmd/main.go
```diff
@@ -0,0 +1,58 @@
...
+	webService.Init()
+
+	obs.InitLogs()
+	wc := restful.NewContainer()
+
+	s3ws := new(restful.WebService)
+	s3ws.Path("/")
+	s3ws.Doc("OpenSDS Multi-Cloud API")
+	s3ws.Consumes(restful.MIME_XML)
+	s3ws.Produces(restful.MIME_XML)
```
sunfch: I dont know the  the reason for their existence of "s3ws.Consumes(restful.MIME_XML) and s3ws.Produces(restful.MIME_XML)",  and when I test s3 compatibility with this pr,  I found "s3ws.Consumes(restful.MIME_XML)" lead to error

s3api/pkg/filters/signature/signer/signature.go
```diff
@@ -0,0 +1,104 @@
...
+			log.Errorf("[%s] reject: IsReqAuthenticated return false, err:%v\n", req.Request.URL, err)
+			s3.WriteErrorResponse(resp, req, err)
+			return
+		}
+		// TODO: check bucket policy
+	case AuthTypeAnonymous:
+		log.Errorf("[%s] reject: Anonymous not supported.\n", req.Request.URL)
+		s3.WriteErrorResponse(resp, req, s3error.ErrSignatureVersionNotSupported)
+		// TODO: check bucket policy
+		return
```
sunfch: i think it is wrong to return directly in the case

docker-compose.yml
sfzeng: Thanks, added.

s3api/cmd/main.go
sfzeng: Could you please share what kind of error?

sfzeng: "Consumes specifies that this WebService can consume one or more MIME types. Http requests must have one of these values set for the Content-Type header."
Which kind of content type you used when error occured?

s3api/pkg/filters/signature/signer/signature.go
sfzeng: Thanks, it's fixed.

s3api/cmd/main.go
sfzeng: As this part is the same as before, and this PR is used to fix the issue that URL and authentication are not compatible aws s3. So can check and fix it in the following PR?

sunfch: http status code 415 was returned when I test s3 compatibility,  I commented "s3ws.Consumes(restful.MIME_XML)"  and the test gone on.  
I agree you that we check the issue in following PR.

sfzeng:
Close it, will raise a new one.

PR #901, opened 2020-03-10
Sses3putobject

rhsakarpos:
this adds support for the S3 header 
image

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # 893
Partly addresses 893

**Release note**:

```
Today, in DAITO release, we support the encryption variant SSE-S3. Headers and values relevant to this are supported by this PR.
Means, if this header and value are specified, the object will be encrypted (even if the bucket does not support encryption)
"x-amz-server-side-encryption": "AES256"
```


rhsakarpos:
@kumarashit @sfzeng @sunfch please review. thanks

api/pkg/filters/signature/main.go
```diff
@@ -0,0 +1,334 @@
+package main
```
kumarashit: Will there be any sync required here after auth v4 changes https://github.com/sodafoundation/multi-cloud/pull/903/files#diff-dca9837476819ec12ee03315c11aca98 either here or in the other code?

sfzeng:
Hi @rhsakarpos , as our signature is changed to support auth v2/v4 by https://github.com/sodafoundation/multi-cloud/pull/903/files#diff-dca9837476819ec12ee03315c11aca98, could you please sync with it?

api/pkg/filters/signature/main.go
rhsakarpos: removed this file

rhsakarpos:
> Please check if binary files, MyRoute.go and main.go files required. It seems some test files are checked in

not needed, removed

s3/pkg/service/object.go
```diff
@@ -197,7 +228,7 @@ func (s *s3Service) PutObject(ctx context.Context, in pb.S3_PutObjectStream) err
 		log.Errorln("failed to create storage. err:", err)
 		return err
 	}
-	obj := &pb.Object{BucketName: req.BucketName, ObjectKey: req.ObjectKey}
```
sfzeng: Hi, why this line is removed ?

PR #907, opened 2020-03-26
Made some change so that lifecycle related APIs can be AWS S3 compatible.

### LINKED ISSUE #893: S3 APIs are not AWS S3 compatible.

sfzeng:
This PR is to make sure lifecylce related APIs can be AWS S3 compatible. The following test cases are passed:
s3tests_boto3.functional.test_s3.test_lifecycle_set
s3tests_boto3.functional.test_s3.test_lifecycle_get
s3tests_boto3.functional.test_s3.test_lifecycle_get_no_id
s3tests_boto3.functional.test_s3.test_lifecycle_id_too_long
s3tests_boto3.functional.test_s3.test_lifecycle_same_id
s3tests_boto3.functional.test_s3.test_lifecycle_invalid_status
s3tests_boto3.functional.test_s3.test_lifecycle_expiration_header_head
s3tests_boto3.functional.test_s3.test_lifecycle_set_noncurrent
s3tests_boto3.functional.test_s3.test_lifecycle_set_multipart

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #893 

It shows there are more than six hundreds of new line, but actually most of them are from protobuf, which is automatically created by protoc.

**Release note**:

```release-note
1. Currently, for expiration rule, we only support expiration days, not support sepcific expiration date.
2. As Prefix in rule is deprecatd, and Filter is suggested for AWS S3 itself. So we only support Filter. And curretnly, we only support Filter:Prefix, not support Filter:Tag.
```


s3/pkg/service/object.go
```diff
@@ -316,11 +318,38 @@ func (s *s3Service) GetObjectMeta(ctx context.Context, in *pb.Object, out *pb.Ge
...
+func getObjectExpDate(lcRules []*pb.LifecycleRule, objKey string, lastModified int64) (int64, string) {
+	var expTime int64 = 0
+	for _, r := range lcRules {
+		if r.Status == utils2.RuleStatusDisabled {
+			continue
+		}
+		for _, ac := range r.Actions {
+			if ac.Name == utils2.ActionNameExpiration {
+				if r.Filter.Prefix == "" || strings.HasPrefix(objKey, r.Filter.Prefix) {
+					exp := lastModified + 60*60*24*int64(ac.Days)
```
kumarashit: Put 60   *  60   *   24 as constant

s3/pkg/service/object.go
```diff
@@ -316,11 +318,38 @@ func (s *s3Service) GetObjectMeta(ctx context.Context, in *pb.Object, out *pb.Ge
...
+					if expTime < exp {
```
kumarashit: When can it land in this situation? i.e. expTime < exp

sfzeng: If there are more than one expiration rule, but expiration days are not the same, we will remove object when the latest expiration time arrived.

s3/pkg/service/object.go
sfzeng: Done

s3/pkg/service/object.go
kumarashit: Got it

PR #908, opened 2020-03-26
fix bug in multipart when resend part

### LINKED ISSUE #916: New repeated item is inserted into db when resend object part for multipart

>
> /kind bug

**What happened**:
New repeated item is inserted into db when resend object part for multipart

Old item should be deleted and new item is inserted into db, and repeated items in objectparts table are not allowed

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:
This pr fix bug that we cannot resend part to s3 server with multipart mode
Old item should be deleted and new item is inserted into db, and repeated items in objectparts table are not allowed

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #916 

**Release note**:

```release-note
```

### DIFF
--- a/s3/pkg/conf/tidb.sql
+++ b/s3/pkg/conf/tidb.sql
@@ -139,7 +139,7 @@ CREATE TABLE IF NOT EXISTS `objectparts` (
   `offset` bigint(20) DEFAULT NULL,
   `etag` varchar(255) DEFAULT NULL,
   `lastmodified` datetime DEFAULT NULL,
-   KEY `rowkey` (`bucketname`,`objectname`,`uploadid`,`partnumber`)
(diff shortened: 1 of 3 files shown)


sfzeng:
Hi @sunfch , there are two functions operate on two database, but only one database table is changed accordingly. Please make sure if that is OK. Thanks.

sunfch:
> Hi @sunfch , there are two functions operate on two database, but only one database table is changed accordingly. Please make sure if that is OK. Thanks.

@sfzeng,  the other table in yig backend already has "unique" key, so in the PR, only one table is modified

PR #910, opened 2020-03-26
fix bug for object copy

### LINKED ISSUE #917: cannot apply acl to target object when copy object

>
> /kind bug

**What happened**:
When send a object copy request with canned acl, the acl cannot be added to the target object

New target object should have the canned acl

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:

There are 3 bugs:
1. The canned acl can not be added to target object in object copy request
2. Update the processing logic of "X-Amz-Metadata-Directive" header.  the object metadata cannot be replaced in object copy request at present.
3. The md5 cannot be returned in put object request
4.  Return nil instead of err in s3 grpc server

The PR fix the bugs as above

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #917 

**Release note**:

```release-note
```


kumarashit:
Hi @sunfch Can you please confirm the following with this fix:
a) As AWS doesn't support Copying ACL in COPY OBJECT, we will now get the ACL from Source and Update it on Target?
b) All possible metadata will now be re-stored to target object because of X-AMZ-METADATA-DIRECTIVE?
c) What is the behaviour for COPY and REPLACE when srcBucketname != DestBucketName?
d) As we are returning nil instead of err, there is no other handling required?

sunfch:
thanks for your review, 

> Hi @sunfch Can you please confirm the following with this fix:
> a) As AWS doesn't support Copying ACL in COPY OBJECT, we will now get the ACL from Source and Update it on Target?
    I dont get the ACL from Source in the PR, and only support a canned ACL in our COPY request using the x-amz-acl request header.
> b) All possible metadata will now be re-stored to target object because of X-AMZ-METADATA-DIRECTIVE?
   According my understanding, if X-AMZ-METADATA-DIRECTIVE is COPY,  metadata from source object will be re-stored to target object, and this is default behavior.
> c) What is the behaviour for COPY and REPLACE when srcBucketname != DestBucketName?
   According my understanding, if X-AMZ-METADATA-DIRECTIVE is COPY, metadata is copied to target object in COPY request, and if if X-AMZ-METADATA-DIRECTIVE is REPLACE, metadata is replaced with the metadata in the COPY request header.
> d) As we are returning nil instead of err, there is no other handling required?
   We define error member in response struct,  and the error is returned to grpc client, and gprc client handle with it

api/pkg/s3/objectcopy.go
```diff
@@ -101,19 +101,6 @@ func (s *APIService) ObjectCopy(request *restful.Request, response *restful.Resp
 		return
 	}
 
-	var isOnlyUpdateMetadata = false
```
sfzeng: Why removed this part?

s3/proto/s3.micro.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by protoc-gen-micro. DO NOT EDIT.
-// source: s3/proto/s3.proto
+// source: s3.proto
```
sfzeng: Hi @sunfch which version of protoc you used, seems the file is renamed.

sunfch: It show version 3.7.1

sfzeng:
> thanks for your review,
> 
> > Hi @sunfch Can you please confirm the following with this fix:
> > a) As AWS doesn't support Copying ACL in COPY OBJECT, we will now get the ACL from Source and Update it on Target?
> 
> I dont get the ACL from Source in the PR, and only support a canned ACL in our COPY request using the x-amz-acl request header.
> 
> > b) All possible metadata will now be re-stored to target object because of X-AMZ-METADATA-DIRECTIVE?
> 
> According my understanding, if X-AMZ-METADATA-DIRECTIVE is COPY, metadata from source object will be re-stored to target object, and this is default behavior.
> 
> > c) What is the behaviour for COPY and REPLACE when srcBucketname != DestBucketName?
> 
> According my understanding, if X-AMZ-METADATA-DIRECTIVE is COPY, metadata is copied to target object in COPY request, and if if X-AMZ-METADATA-DIRECTIVE is REPLACE, metadata is replaced with the metadata in the COPY request header.
> 
> > d) As we are returning nil instead of err, there is no other handling required?
> 
> We define error member in response struct, and the error is returned to grpc client, and gprc client handle with it

Hi @sunfch , does that mean, if X-Amz-Metadata-Directive is set to be COPY, then all metadata of target object will be set the same as source object, except ACL, and if X-Amz-Metadata-Directive is set to be REPLACE, then all of targe object will be set according to the request?

sunfch:
> > thanks for your review,
> > > Hi @sunfch Can you please confirm the following with this fix:
> > > a) As AWS doesn't support Copying ACL in COPY OBJECT, we will now get the ACL from Source and Update it on Target?
> > 
> > 
> > I dont get the ACL from Source in the PR, and only support a canned ACL in our COPY request using the x-amz-acl request header.
> > > b) All possible metadata will now be re-stored to target object because of X-AMZ-METADATA-DIRECTIVE?
> > 
> > 
> > According my understanding, if X-AMZ-METADATA-DIRECTIVE is COPY, metadata from source object will be re-stored to target object, and this is default behavior.
> > > c) What is the behaviour for COPY and REPLACE when srcBucketname != DestBucketName?
> > 
> > 
> > According my understanding, if X-AMZ-METADATA-DIRECTIVE is COPY, metadata is copied to target object in COPY request, and if if X-AMZ-METADATA-DIRECTIVE is REPLACE, metadata is replaced with the metadata in the COPY request header.
> > > d) As we are returning nil instead of err, there is no other handling required?
> > 
> > 
> > We define error member in response struct, and the error is returned to grpc client, and gprc client handle with it
> 
> Hi @sunfch , does that mean, if X-Amz-Metadata-Directive is set to be COPY, then all metadata of target object will be set the same as source object, except ACL, and if X-Amz-Metadata-Directive is set to be REPLACE, then all of targe object will be set according to the request?

@sfzeng ,  yes, this is my understand too

kumarashit:
Thanks for the explanation. Yes, when i asked about X-AMZ-Metadata-Directive for copy, i meant all except ACL.

PR #911, opened 2020-03-26
fix bug for multipart

### LINKED ISSUE #918: EntityTooSmall is not returned when object part size is less than 5MB

>
> /kind bug

**What happened**:
According to aws doc,  when object part is less 5MB except last part, EntityTooSmall should be returned

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunfch:

There are some bugs:
1. Cannot return EntityTooSmall when object part size is less than 5M for multipart
2. In S3 grpc server, nil should be returned instead err

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #918 

**Release note**:

```release-note
```


s3/pkg/service/multipart.go
```diff
@@ -384,8 +385,15 @@ func (s *s3Service) CompleteMultipartUpload(ctx context.Context, in *pb.Complete
 		if part.Etag != in.CompleteParts[i].ETag {
 			log.Errorln("part etag in meta store is not the same with client's part, partno:", i)
 			err = ErrInvalidPart
-			return err
```
sfzeng: Hi @sunfch Why donnot check size in the UploadPart API, but check it in the CompleteMultipart API?

sunfch: because object part size should be less 5MB except last part for multipart

sfzeng: OK, does AWS S3 do it in the same way?

sunfch: Yes,  pls refer to  https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html

kumarashit: > @sfzeng , because object part size should be less 5MB except last part for multipart

You mean, should not be less tha 5MB except last part can be less than 5MB. right?

sunfch: @kumarashit ,  yes, the last part is allowed to be less than 5MB

PR #914, opened 2020-03-26
fix bug for s3 compatibility

sunfch:

1. return nil in s3 grpc server
2. add code to check the validity of md5 in the http header
3. return the value of "cache-control" from object metadata
4. return ErrNotImplemented for versioning request

**Release note**:

```release-note
```


api/pkg/s3/object-utils.go
```diff
@@ -75,10 +75,6 @@ func isValidObjectName(objectName string) bool {
 		if (n >= 0 && n <= 31) || (n >= 127 && n <= 255) {
 			return false
 		}
-		c := string(n)
```
kumarashit: Reason to remove this?

sunfch: because in s3-tests, some test case that object name contain the character “\” cannot pass.  I dont sure that it is correct to remove this.  @sfzeng  pls help review this too, thanks

sfzeng: Hi @sunfch did that test pass for AWS S3? It seems backslash('\') is invalid for aws s3, please check https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys

sunfch: @sfzeng , I upload a object whose name contain backslash character successfully,  and in the url link, it show the character should avoid, I also not sure and confused with that

kumarashit: It seems the backslash should be avoided but AWS doesn't reject it outright. To be on safer side let's fix this not to allow '\'

sunfch: @kumarashit , Yes, I dont sure that the code snippet should be removed.  @sfzeng , what about your opinion

sfzeng: If we removed this part, we have to test and make sure we can list/get/copy this object with this special name. Can we make sure? @sunfch

sunfch: Because AWS suggest that the backslash should be avoided,  we dont remove the code snippet for the present？

sfzeng: How about, we do not allow backslash, and declare it in the release note?

sunfch: It depends on your decision.  If you want to remove it I will test list/get/copy object with the backslash.

sunfch: @sfzeng  I have tested the object with this special name for list/get/copy/put api, and the successful results are returned.

PR #966, opened 2020-05-22
To fix issue of downloading migrated object (Size <5MB)

### LINKED ISSUE #952: Unable to download Migrated objects from destination buckets

>
kind bug
> /kind feature

**What happened**:
After migrating object from source bucket to destination bucket, user will not be able to download objects from destination buckets

User should be able to download objects from destination buckets.

1. Install Gelato project with `development` branch
2. Create backends and create 2 buckets (i.e. source and destination)
3. Upload objects to Source bucket
4. Migrate objects ( Objects got migrated successfully and can be downloaded from cloud console)
5. Try to download objects using OpenSDS dashboard or API [FAILED]

**Environment**:
- Gelato(release/branch) version: development (Elba release)
- Kernel (e.g. `uname -a`):
- Install tools: 
- Others: 

S3-API logs

image

S3 service logs

image

TiDB data
image

Click2cloud-Kukekar:
**What type of PR is this?**
>
> /kind new feature

bug fix
> /kind cleanup
> /kind revert change
> /kind design
> /kind documentation
> /kind enhancement

To fix the issue of downloading migrated objects( size <5MB) , as earlier the ACL properties were not updated in tidb hence it was getting `null` value for ACL while downloading object. Even userId was updated in database while migrating object, so added that part too.

Fixes #952 

**Test Report Added?**:

TESTED
> /kind NOT-TESTED

**Test Report**:


s3/pkg/service/object.go
```diff
@@ -630,8 +630,9 @@ func (s *s3Service) CopyObject(ctx context.Context, in *pb.CopyObjectRequest, ou
 	targetObject.StorageMeta = res.Meta
 	targetObject.Location = targetBackendName
 	targetObject.TenantId = tenantId
+	targetObject.UserId = srcBucket.UserId
 	// this is the default acl setting
-	targetObject.Acl = in.Acl
+	targetObject.Acl = targetBucket.Acl
```
sfzeng, line 635: Hi @Click2cloud-Kukekar , why use targetObject.Acl as the targetBucket.Acl, IMO, user can specify Acl in the copy request.

Click2Cloud-Gamma: @sfzeng Okay, but are we giving any option (on dashboard) to user to add ACL details while copy object operation. If yes, then we have to put 1 condition when this function will be used for migration purpose.
During migration in.ACL was `null` which create issue while downloading object from target bucket as at ACL checkpoint it gets `null` value.

sfzeng: On dashboard there is no such option, but it can be used through REST request directly, so can we set it to be targetBucket.Acl if in.ACL is not set, as it's the default config, and if in.ACL is set then we use it?

Click2Cloud-Gamma: @sfzeng Made changes as suggested, please review.
Thanks

s3/pkg/service/object.go
```diff
@@ -630,8 +630,13 @@ func (s *s3Service) CopyObject(ctx context.Context, in *pb.CopyObjectRequest, ou
 	targetObject.StorageMeta = res.Meta
 	targetObject.Location = targetBackendName
 	targetObject.TenantId = tenantId
+	targetObject.UserId = srcBucket.UserId
```
sfzeng, line 633: Here, userid can be get from ctx, see line 521, not from source bucket.

Click2Cloud-Gamma: @sfzeng Done , Thanks

PR #1236, opened 2021-03-30
Fix Block and File backend registration

### LINKED ISSUE #1235: Cloud Block and File backend registration fails
**Issue/Feature Description:**
Cloud Block and File backend registration fails with backend check failure

**Why this issue to fixed / feature is needed(give scenarios or use cases):**
Blocking for File and Block service functionality

**How to reproduce, in case of a bug:**
Register backend by choosing block/file. It will fail

**Other Notes /  Environment Information: (Please give the env information, log link or any useful information for this issue)**

kumarashit:
**What type of PR is this?**
>
> /kind new feature
/kind bug fix
> /kind cleanup
> /kind revert change
> /kind design
> /kind documentation
> /kind enhancement

This PR fixes the issue of backend registration for File and Block. This is must fix
The BackendCheck() function is very specific to S3. So this should be called only for object storage type.
Added a check there.

Fixes #1235 

**Test Report Added?**:
/kind TESTED
> /kind NOT-TESTED

**Test Report**:
image
image
image

Apart from this tested:

- [x] Azure Blob Registration
- [x] AWS S3 Registartion
- [x] Cloud bucket create 
- [x] Object Upload

### DIFF
--- a/api/pkg/backend/service.go
+++ b/api/pkg/backend/service.go
@@ -25,7 +25,6 @@ import (
 	c "github.com/opensds/multi-cloud/api/pkg/context"
 	"github.com/opensds/multi-cloud/api/pkg/filters/signature/credentials/keystonecredentials"
 	"github.com/opensds/multi-cloud/api/pkg/policy"
-	"github.com/opensds/multi-cloud/api/pkg/utils"
 	"github.com/opensds/multi-cloud/api/pkg/utils/cryptography"
 	backend "github.com/opensds/multi-cloud/backend/proto"
 	dataflow "github.com/opensds/multi-cloud/dataflow/proto"
@@ -48,8 +47,20 @@ const (
 	dataflowService_K8S    = "soda.multicloud.v1.dataflow"
 )
 
-// List of object storge providers supported by s3 services
-var object_types = []string{"aws-s3", "azure-blob", "ibm-cos", "hw-obs", "ceph-s3", "gcp-s3", "fusionstorage-object", "yig", "alibaba-oss", "sony-oda"}
+// Map of object storge providers supported by s3 services. Keeping a map
+// to optimize search in o(1)
+var objectStorage = map[string]int{
+	"aws-s3":               1,
+	"azure-blob":           1,
+	"ibm-cos":              1,
+	"hw-obs":               1,
+	"ceph-s3":              1,
+	"gcp-s3":               1,
+	"fusionstorage-object": 1,
+	"yig":                  1,
+	"alibaba-oss":          1,
+	"sony-oda":             1,
+}
 
 type APIService struct {
 	backendClient  backend.BackendService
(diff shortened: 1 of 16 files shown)


api/pkg/backend/service.go
```diff
@@ -47,6 +48,9 @@ const (
 	dataflowService_K8S    = "soda.multicloud.v1.dataflow"
 )
 
+// List of object storge providers supported by s3 services
+var object_types = []string{"aws-s3", "azure-blob", "ibm-cos", "hw-obs", "ceph-s3", "gcp-s3", "fusionstorage-object", "yig", "alibaba-oss", "sony-oda"}
```
AmitRoushan, line 52: var objectStoarge = map[string]struct{}{
                                   "aws-s3": struct{}{},
                                    "azure-blob": struct{}{},
                                    "ibm-cos": struct{}{}, 
                                    "hw-obs": struct{}{},
                                     "ceph-s3": struct{}{},
                               }
func isObjectStorage(storage string) bool {
     _, ok := objectStoarge[storage ]
     return ok
}

IMHO, Above might be better implementation

AmitRoushan (changes requested):

api/pkg/backend/service.go
```diff
@@ -47,6 +48,9 @@ const (
 	dataflowService_K8S    = "soda.multicloud.v1.dataflow"
 )
 
+// List of object storge providers supported by s3 services
```
anvithks, line 51: nit: object storge - > object storage

api/pkg/utils/utils.go
```diff
@@ -215,3 +215,13 @@ func GetTimeoutSec(objSize int64) int64 {
 
 	return tmoutSec
 }
+
+// Function to check if the slice contains a string
+func SliceContainsString(inp string, list []string) bool {
```
PravinRanjan10, line 220: May be function can be renamed. ex: ObjectTyeStorageCheck(), this will give fair idea, during readability of code

PravinRanjan10:
LGTM

api/pkg/backend/service.go
kumarashit: Thanks for the suggestion. Yes optimized it now. o(n) -> o(1) :)

api/pkg/backend/service.go
```diff
@@ -246,12 +250,15 @@ func (s *APIService) CreateBackend(request *restful.Request, response *restful.R
...
-	_, err = s.s3Client.BackendCheck(ctx, backendDetailS3)
-	err1 := errors.New("Failed to register backend due to invalid credentials.")
-	if err != nil {
-		log.Errorf("failed to create backend due to wrong credentials: %v", err)
-		response.WriteError(http.StatusBadRequest, err1)
-		return
+    // This backend check will be called only for object storage
+	if utils.SliceContainsString(backendDetail.Type, object_types) {
+		_, err = s.s3Client.BackendCheck(ctx, backendDetailS3)
+		err1 := errors.New("Failed to register backend due to invalid credentials.")
```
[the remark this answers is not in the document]
kumarashit: Moved it inside the loop

api/pkg/backend/service.go
kumarashit: done

AmitRoushan (approved):
LGTM

NajmudheenCT (approved):
LGTM

skdwriting (approved):
LGTM

PR #1281, opened 2021-06-27
Json Formatting for AKSK Response 

### LINKED ISSUE #1279: Can the response of the Create / Get AK/SK API be a JSON value instead of a string?
**Issue/Feature Description:**
As per the current implementation of the AK/SK API, the response is as follows:

get-ak-sk

The response is inline with how Keystone returns the Credentials, where the Ak/SK is a string object
```
{
 ...
      "blob": "{\"access\":\"wuGNnh8ZaOl2vRMw\",\"secret\":\"2pMBb8J18Lrrk27baw9hKAZsGxLYahbi\"}",
 ...
}

```

can this be changed to a simple JSON object with key value pairs instead?
Something like:
```
{
  ...
    access_key: "wuGNnh8ZaOl2vRMw",
    secret_key: "2pMBb8J18Lrrk27baw9hKAZsGxLYahbi"
  ...
}
```

**Why this issue to fixed / feature is needed(give scenarios or use cases):**
The current implementation follows the keystone response structure. Since we do not have control over the Keystone response we use the values as available . This means the client has to parse this response, convert the string to a valid JSON and only then is able to use the resultant values.
Since we are now providing an API that abstracts all of this, it will help if the client can just consume the API and fetch the credentials as  a JSON.

One caveat is that having the AK/SK in plain text is not really ideal. Maybe we can use the token and have the AK/SK as part of the token or use some encryption mechanism and ensure that the client uses the same to decrypt the credentials.

**How to reproduce, in case of a bug:**

**Other Notes /  Environment Information: (Please give the env information, log link or any useful information for this issue)**

rajat-soda:
**What type of PR is this?**
Provides AKSK Response in JSON format  , for easier consumption by 3rd party Clients. 

>
kind enhancement

Provides AKSK Response in JSON format  , for easier consumption by 3rd party Clients. 

Fixes #1279

**Test Report Added?**:
kind TESTED

**Test Report**:

image
image


rajat-soda:
@PravinRanjan10 , @anvithks , @AmitRoushan , @kumarashit  Please review

kumarashit (approved):
LGTM

rajat-soda:
@kumarashit :  adding formatting for the GET API aswell. hold on for some time.

skdwriting (approved):
LGTM

PR #1360, opened 2022-11-26
Starto migration status not updating fix and Backend listing as SSP fix

josephb160515cs:
**What type of PR is this?**
/kind bug fix

1) In Strato functional testing yesterday discovered that after creating migration plans. Even though object is migrated from src to the target bucket the status of the migration plan doesn't reflect the same due to an internal issue in code related to the database update

2) In Storage Service Plan mode of installation backends was getting listed as SSPs

Fixes #

**Test Report Added?**:
/kind TESTED

**Test Report**:
#### Storage Service Plan Testing
Creating a service plan now lists the service plan instead of the previous backends 
image

Creating bucket with SSP is now working
image

#### Migration from one cloud vendor to another
- Migration from AWS to GCP
Status and number of objects migrated is correct
image
Object is also present in the target bucket 
image

- Migration from GCP to AWS
Status and number of objects migrated is correct
image
Object is aslo present in the target bucket
image

### DIFF
--- a/backend/pkg/db/drivers/mongo/mongo.go
+++ b/backend/pkg/db/drivers/mongo/mongo.go
@@ -257,7 +257,7 @@ func (repo *mongoRepository) ListTiers(ctx context.Context, limit, offset int, q
 	}
 	log.Infof("ListTiers, limit=%d, offset=%d, m=%+v\n", limit, offset, m)
 
-	cur, err := session.Database(defaultDBName).Collection(defaultCollection).Find(ctx, m)
+	cur, err := session.Database(defaultDBName).Collection(defaultTierCollection).Find(ctx, m)
 
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/hashicorp/vault/api v1.0.5-0.20190814205542-3b036e58e950
```
PravinRanjan10, line 25: no need to update this file. plz Don't add to this PR

go.sum
```diff
@@ -350,6 +350,8 @@ github.com/jmespath/go-jmespath v0.3.0/go.mod h1:9QtRXoHjLGCJ5IBSaohpXITPlowMeeY
 github.com/jonboulle/clockwork v0.1.0/go.mod h1:Ii8DK3G1RaLaWxj9trq07+26W01tbo22gdxWY5EU2bo=
```
PravinRanjan10, line 350: no need to update this file. plz Don't add to this PR

PravinRanjan10 (changes requested):

go.mod
josephb160515cs: added temporarily for testing , will revert that commit after same

go.sum
josephb160515cs: added temporarily for testing , will revert that commit after same

kumarashit (approved):
LGTM

PravinRanjan10 (approved):

PR #1368, opened 2022-12-15
install docker-py package before checking for keystone installation

devanshjain7:
**What type of PR is this?**
/kind bug fix

checking for keystone installation requires docker-py package. Hence, it should be installed before the task where checking for keystone installation happens.


joseph-v (approved):
LGTM

skdwriting (approved):
LGTM

PR #1374, opened 2022-12-28
Metadata Manager : List Metadata API enhancement as per latest design docs

josephb160515cs:
**What type of PR is this?**
> /kind enhancement

CurrenListMetadataAPI doesnt support almost all the request parameters mentioned in the design docs.
This PR fetches the request parameters from the request from the api handler and sends to metadata manager m8s using gRPC

Fixes #

**Test Report Added?**:
> /kind TESTED
* Limit & Offset is valid 
image

* Limit/Offset is invalid
image


api/pkg/metadata/service.go
```diff
@@ -68,3 +74,47 @@ func (s *APIService) ListMetadata(request *restful.Request, response *restful.Re
...
+	offsetStr := request.PathParameter("offset")
+	bucketName := request.PathParameter("bucketName")
+	objectName := request.PathParameter("objectName")
+	sizeOfObject := request.PathParameter("sizeOfObject")
+	sizeOfBucket := request.PathParameter("sizeOfBucket")
+
+	offsetStr, limitStr = assignDefaultValues(offsetStr, limitStr)
+
+	// convert string offset to int32
+	offset, err := strconv.ParseInt(offsetStr, 10, 32)
```
PravinRanjan10, line 93: You may use  GetPaginationParam() function from common.go

api/pkg/metadata/service.go
```diff
@@ -68,3 +74,47 @@ func (s *APIService) ListMetadata(request *restful.Request, response *restful.Re
...
+	if err != nil {
+		log.Errorf("Failed to parse offset as int32: \n", err)
+		return mt.ListMetadataRequest{}, err
+	}
+
+	// convert string limit to int32
+	limit, err := strconv.ParseInt(limitStr, 10, 32)
```
PravinRanjan10, line 100: same as above comment

metadata/proto/metadata.proto
```diff
@@ -18,7 +18,23 @@ message SyncMetadataRequest {
...
+    string backendName = 4;
+    string region = 5;
+    // name of the bucket
+    string bucketName = 6;
+    // name of the object
+    string objectName = 7;
+    // the size of the object in <operator integer  unit> where operator is eq, gte,gt,lte,lt and unit is KB,MB,GB and integer is always not less than zero
+    string sizeOfObject = 8;
+    // the size of the bucket in <operator integer  unit> where operator is eq, gte,gt,lte,lt and unit is KB,MB,GB and integer is always not less than zero
+    string sizeOfBucket = 9;
```
PravinRanjan10, line 37: why string?, better to add as Integer.

metadata/proto/metadata.proto
```diff
@@ -18,7 +18,23 @@ message SyncMetadataRequest {
...
+    string type = 3;
+    //name of the backend
+    string backendName = 4;
+    string region = 5;
+    // name of the bucket
+    string bucketName = 6;
+    // name of the object
+    string objectName = 7;
+    // the size of the object in <operator integer  unit> where operator is eq, gte,gt,lte,lt and unit is KB,MB,GB and integer is always not less than zero
+    string sizeOfObject = 8;
```
PravinRanjan10, line 35: why string?, better to add as Integer.

PravinRanjan10 (changes requested):

api/pkg/metadata/service.go
josephb160515cs: done

metadata/proto/metadata.proto
josephb160515cs: done

api/pkg/metadata/service.go
josephb160515cs: done

metadata/proto/metadata.proto
josephb160515cs: As per discussion converted to integer and added another field for operation

metadata/proto/metadata.proto
josephb160515cs: As per discussion converted to integer and added another field for operation

api/pkg/metadata/service.go
```diff
@@ -51,20 +51,55 @@ func (s *APIService) SyncMetadata(request *restful.Request, response *restful.Re
...
+//* It returns ListMetadataRequest for ListMetaData API call
+func GetListMetaDataRequest(request *restful.Request) (listMetadataRequest mt.ListMetadataRequest, err error) {
+	typeOfCloudVendor := request.PathParameter("type")
+	backendName := request.PathParameter("backendName")
+	bucketName := request.PathParameter("bucketName")
+	objectName := request.PathParameter("objectName")
+	sizeOfObjectInBytes, err := common.GetSizeRequestParamAsInt64(request, "sizeOfObject")
+
+	if err != nil {
+		return mt.ListMetadataRequest{}, err
```
PravinRanjan10, line 89: Please avoid returing err messge directly to user. You may return.

type of sizeOfObject is invalid. It should be Integer type.

josephb160515cs: Done

PravinRanjan10 (approved):
LGTM

PR #1380, opened 2023-01-11
[metadata manager m8s] [Validator] implementation for validator

josephjacobmorris:
**What type of PR is this?**
> /kind feature

This PR helps in validating different query params . and  gives appropriate message to the user if it is invalid. 

**Test Report Added?**:
> /kind TESTED

**Test Report**:
* Valid Scenario
image

* If region is invalid and type is given as AWS-S3
image

* If SizeOfBucket is invalid
image

* if Size Operator is invalid
image


metadata/pkg/service/service.go
```diff
@@ -157,7 +158,18 @@ func (f *metadataService) SyncMetadata(ctx context.Context, in *pb.SyncMetadataR
 }
 
 func (f *metadataService) ListMetadata(ctx context.Context, in *pb.ListMetadataRequest, out *pb.ListMetadataResponse) error {
-	log.Info("Received GetMetadata request in metadata service.")
+	log.Info("received GetMetadata request in metadata service.")
```
PravinRanjan10, line 161: plz be cosistent with either starting char should be caps or small.

metadata/pkg/validator/AWSQueryValidator.go
```diff
@@ -0,0 +1,31 @@
+package validator
```
PravinRanjan10, line 1: just to validate region, why do we need seprate file.

PravinRanjan10, line 1: just to validate region, why do we need seprate file.

metadata/pkg/validator/generalQueryValidator.go
```diff
@@ -0,0 +1,48 @@
...
+
+	case constants.LESS_THAN_EQUAL_OPERATOR:
+
+	case constants.LESS_THAN_OPERATOR:
+
+	case constants.GREATER_THAN_EQUAL_OPERATOR:
+
+	case constants.GREATER_THAN_OPERATOR:
+		// operator is valid
+		break
```
PravinRanjan10, line 38: why do we need break here?

metadata/pkg/constants/Constants.go
```diff
@@ -0,0 +1,36 @@
+package constants
+
+const (
+	CACHE_NAME = "cachename"
+
```
PravinRanjan10, line 5: you may remove gap of line between two lines.

PravinRanjan10 (changes requested):

metadata/pkg/service/service.go
josephjacobmorris: keeping all info logs in lower case 
keeping all error logs starting with uppercase

metadata/pkg/validator/AWSQueryValidator.go
josephjacobmorris: moved all validation logic to validator file

metadata/pkg/validator/generalQueryValidator.go
josephjacobmorris: moved all validation logic to validator file

metadata/pkg/constants/Constants.go
josephjacobmorris: done

metadata/pkg/validator/AWSQueryValidator.go
josephjacobmorris: moved all validation logic to validator file

PravinRanjan10 (approved):
LGTM

PR #1395, opened 2023-02-03
aws driver updated to collect bucket metadata associated with backend region only

devanshjain7:
**What type of PR is this?**
/kind enhancement

Currently metadata collection is being done for all the buckets irrespective of the region. But now as per the requirement, only buckets associated with the backend region need to be considered for metadata collection. This PR updates the aws driver for fulfilling the requirement.

### DIFF
--- a/metadata/pkg/drivers/aws/aws.go
+++ b/metadata/pkg/drivers/aws/aws.go
@@ -20,7 +20,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
@@ -51,12 +50,12 @@ type AwsAdapter struct {
 	Session *session.Session
 }
 
-func ObjectList(sess *session.Session, bucket *model.MetaBucket) {
-	svc := s3.New(sess, aws.NewConfig().WithRegion(bucket.Region))
+func ObjectList(sess *session.Session, bucket *model.MetaBucket) error {
+	svc := s3.New(sess)
 	output, err := svc.ListObjectsV2(&s3.ListObjectsV2Input{Bucket: &bucket.Name})
 	if err != nil {
 		log.Errorf("unable to list objects in bucket %v. failed with error: %v", bucket.Name, err)
-		return
+		return err
 	}
 
 	numObjects := len(output.Contents)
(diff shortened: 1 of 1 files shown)


metadata/pkg/drivers/aws/aws.go
```diff
@@ -99,43 +98,46 @@ func ObjectList(sess *session.Session, bucket *model.MetaBucket) {
...
 		}
 		buck.BucketTags = tagset
 	}
-
-	ObjectList(sess, buck)
-	if err != nil {
-		return err
-	}
+	bucketArray[buckIdx] = buck
+	*numBuckets += 1
```
josephjacobmorris, line 140: may cause race condition

metadata/pkg/drivers/aws/aws.go
```diff
@@ -99,43 +98,46 @@ func ObjectList(sess *session.Session, bucket *model.MetaBucket) {
...
+	if err != nil {
+		return err
 	}
 
 	tags, err := svc.GetBucketTagging(&s3.GetBucketTaggingInput{Bucket: bucket.Name})
 
 	if err != nil && !strings.Contains(err.Error(), "NoSuchTagSet") {
 		log.Errorf("unable to get bucket tags. failed with error: %v", err)
-	} else {
+	} else if err == nil {
```
josephjacobmorris, line 132: can refactor to  if err == nil {} else  if !strings.Contains(err.Error(), "NoSuchTagSet")

metadata/pkg/drivers/aws/aws.go
```diff
@@ -99,43 +98,46 @@ func ObjectList(sess *session.Session, bucket *model.MetaBucket) {
 	bucket.NumberOfObjects = numObjects
 	bucket.TotalSize = totSize
 	bucket.Objects = objectArray
+	return nil
 }
 
-func GetBucketMeta(buckIdx int, bucket *s3.Bucket, sess *session.Session, bucketArray []*model.MetaBucket, wg *sync.WaitGroup) error {
+func GetBucketMeta(buckIdx int, bucket *s3.Bucket, sess *session.Session, bucketArray []*model.MetaBucket, numBuckets *int, wg *sync.WaitGroup) error {
```
josephjacobmorris, line 104: give meaningful variable names and avoid naming them as i etc

josephjacobmorris, line 104: I feel no need for this return value

josephjacobmorris (changes requested):

metadata/pkg/drivers/aws/aws.go
devanshjain7: yes thanks for pointing out. its resolved now.

metadata/pkg/drivers/aws/aws.go
devanshjain7: done

metadata/pkg/drivers/aws/aws.go
devanshjain7: done

metadata/pkg/drivers/aws/aws.go
devanshjain7: done

PravinRanjan10 (approved):
LGTM

PR #1405, opened 2023-02-15
[metadata manager m8s] Cache Impl for List Metadata

josephjacobmorris:
**What type of PR is this?**

> /kind new feature
> /kind enhancement

This PR implements the result caching feature for List Metadata Service which will improve the query performance on subsequent same queries to the API within the TTL (Time to Live).

**Test Report Added?**:
> /kind TESTED

**Test Report**:
- Logs showing that result is stored/fetched from cache
image

- without cache
withoutcache

- with cache
cached


josephjacobmorris (commented):
Self Review Done

metadata/pkg/db/drivers/keydb/keydb.go
```diff
@@ -0,0 +1,107 @@
...
+
+func ExitCache() {
+	adap.session.Close()
+}
+
+type Adapter struct {
+	session *redis.Client
+}
+
+func (a Adapter) StoreData(ctx context.Context, result interface{}, query interface{}) error {
```
devanshjain7, line 64: better to use pointer to Adapter while defining StoreData and FetchData methods.

josephjacobmorris: Not required I feel

PR #1406, opened 2023-02-15
add s3 interface for s3 client mocking

devanshjain7:
**What type of PR is this?**
> /kind enhancement

Added s3 interface to enable mocking of the aws s3 client. This will make the unit testing of aws storage driver easier and faster as it will avoid making actual API calls to the aws server.


josephjacobmorris (approved):
LGTM

josephjacobmorris (approved):
LGTM

PR #1409, opened 2023-02-21
vendor files updated using "go mod vendor"

josephjacobmorris:
**What type of PR is this?**

> /kind cleanup

Since we are tracking the vendor files with our version control those also need to updated incase of dependency change which was not done for radoshttp

Fixes #

**Test Report Added?**:
> /kind TESTED


PravinRanjan10 (approved):
LGTM

PR #1410, opened 2023-03-06
support azure driver for metadata management

devanshjain7:
**What type of PR is this?**
> /kind new feature

This PR adds support for Azure Blob Storage for metadata collection.

**Please Note:**
Currently not able to collect the following object level parameters:
Tags, ACL, Expires Date, Encryption, VersionID
**Reason**:  Due to usage of old version of Azure SDK (azure-storage-blob-go@v0.8.0)

**Test Report Added?**:
> /kind TESTED

**Test Report**:
image
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,9 +3,9 @@ module github.com/opensds/multi-cloud
 go 1.17
 
 require (
-	github.com/Azure/azure-pipeline-go v0.2.2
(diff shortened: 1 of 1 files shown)


PravinRanjan10 (changes requested):
Please separate the vendor pkg out of this PR

